Issue #5, opened 2022-09-23
Enhance middle machine to enable dynamic topologies


klingaard:
Currently the olympia model has 1 ALU, 1 branch, etc and Dispatch is hard-coded to this topology.  Using Sparta's extension support, add a dynamic topology definition to allow a user to create different topologies.

This includes (as a start):

1. Defining the number of ALUs, branch units, load/store units, etc
2. Defining structures like instruction issuing queues and pipes (and the ratios for them)

knute-sifive:
Work is complete.

PR #18, opened 2022-11-26
 Issue #5: Dynamic Execution Topologies

knute-sifive:
In this PR, Execute is now it's own node and dynamically creates execution pipes based on extensions listed in the
config files:
```
top.cpu.core0.extension.core_extensions:
  execution_topology:
  [["alu", "6"],
   ["fpu", "2"],
   ["br",  "2"]]
```

Work done;
- Added new class `Dispatcher` that is dynamically created by `Dispatch` for each execution pipe defined in the `execution_topology` extension
- Added new class `Execute` that is a node created for the sole purpose of dynamically creating the `ExecutePipe` objects based on the `execution_topology` extension
- Cleaned up old code
- Reduced the `CPUTopology` structure to just static components like Fetch, Decode, Rename, Dispatch, Execute, LS/etc, and Retire
- Updated the arches to include a small, medium, and big core.  Dhrystone trace improves with each architecture, but is limited by LS.


knute-sifive:
closes #5

core/Execute.cpp
```diff
@@ -1,163 +1,69 @@
...
+            const std::string disp_credits_in = "dispatch.ports.in_" + exe_pipe_rtn->getName() + "_credits";
+            bind_ports(exe_credits_out, disp_credits_in);
+
+            const std::string exe_inst_in   = "execute." + exe_pipe_rtn->getName() + ".ports.in_execute_write";
+            const std::string disp_inst_out = "dispatch.ports.out_" + exe_pipe_rtn->getName() + "_write";
+            bind_ports(exe_inst_in, disp_inst_out);
+
+            const std::string exe_flush_in = "execute." +  exe_pipe_rtn->getName() + ".ports.in_reorder_flush";;
+            const std::string flush_manager = "flushmanager.ports.out_retire_flush";
+            bind_ports(exe_flush_in, flush_manager);
```
aka2en, line 66: Hm, this is something I'm always wondering about -- bind these units here or at topology level? One concern I come up with is -- that would make the execution unit test have to create dispatch instance.

sim/main.cpp
```diff
@@ -90,7 +93,7 @@ int main(int argc, char **argv)
 
         // Create the simulator
         sparta::Scheduler scheduler;
-        OlympiaSim sim("core_topology_1",
+        OlympiaSim sim("core_topology_simple",
```
aka2en, line 96: I'm lazy -- I like fewer words ...
```suggestion
        OlympiaSim sim("simple",
```

aka2en (approved):
LGTM

core/Execute.cpp
knute-sifive: This is true -- I do "hard-code" dispatch and the flush manager in this factory.  I don't like that either.  Perhaps moving this to the topology would make sense...

knute-sifive: Done -- moved to CPUTopology.

PR #22, opened 2022-11-27
Add logger macros to make our lives easier

aka2en:


core/LogUtils.hpp
```diff
@@ -0,0 +1,21 @@
+// <LogUtils.hpp> -*- C++ -*-
```
klingaard, line 1: This file is already in Sparta:  sparta/utils/LogUtils.hpp

klingaard (approved):
I'll approve to get this rolling, but move to the sparta version of the LogUtils.hpp instead.

core/LogUtils.hpp
aka2en: Ok

Issue #2, opened 2022-09-23
Need operand dependency checking


klingaard:
Olympia has a rename stage, but it's pass through. Need some form of a scoreboard (like the one proposed https://github.com/sparcians/map/issues/263) to keep track of dependencies. The faster, the better.

knute-sifive:
Ready to close, but one more item left to do: remove the Limitations section in README.md.  Thanks @AaronGChan !  And great work.

PR #40, opened 2023-03-06
Issue #2: Need Operand Dependency Checking Task 1

AaronGChan:
Summary of Changes:
Rename -> added mapping from ARF -> PRF, freelist, and retirement processing
Rename_test -> created a modified sink unit, test runs a .json file of instructions with various dependencies introduced. The test runs 2 simulations stopping after a certain amount of ticks to check if the freelist, map_table, and reference_counters are being updated correctly.
Dispatch -> modified import of WeightedContextCounter.hpp to include the updated pathing in sparta/statistics

Summary of Work to do for Task 1:
Add a arch map -> PRF map to rename.  This includes adding a physical register freelist (parameterized to the number).  When an instruction enters rename, its source arch registers are mapped to the current physical register for that arch value.  Then its destination registers are renamed — a new PRF is pulled from the freelist, given to the instruction (via the scoreboard mask) and the rename map updated.  Suggest write a Rename_test (copy Dispatch_test in test/core), set up expected scenarios and test your changes.  When this is working, mark all registers ready (so that the rest of the core still works)


core/Rename.cpp
```diff
@@ -10,7 +10,7 @@
 #include "sparta/events/StartupEvent.hpp"
 #include "sparta/app/FeatureConfiguration.hpp"
 #include "sparta/report/DatabaseInterface.hpp"
-
+#include "mavis/ExtractorDirectInfo.h"
```
knute-sifive, line 13: Why do you need to include this header?

core/Rename.cpp
```diff
@@ -83,6 +86,11 @@ namespace olympia
             scoreboards_[rf]->set(bits);
         }
 
+        // initialize freelist
+        for(unsigned int i = 0; i < (int) sparta::Scoreboard::MAX_REGISTERS; i++){
+            freelist[core_types::RegFile::RF_INTEGER].push(i); // INTEGER
+            freelist[core_types::RegFile::RF_FLOAT].push(i); // FLOAT
+        }
```
knute-sifive, line 93: The number of free renames should come from a parameter, not dictated by the size of the scoreboard.  Most hardware designs have a limited number of renames.   

Also, it's not just 1 parameter -- it's a parameter per register file.  For example, there should be one for integer renames, one for float-point renames, etc.

Finally, when you add the parameters, you'll need to break up this `for` loop into each `RegFile` type.  I'd write a `lamda` function that would do this for me.

core/Rename.cpp
```diff
@@ -96,7 +104,59 @@ namespace olympia
             ev_rename_insts_.schedule();
         }
     }
+    void Rename::getAckFromROB_(const InstPtr & inst_ptr)
+    {
+        sparta_assert(inst_ptr->getStatus() == Inst::Status::RETIRED,
+                        "Get ROB Ack, but the inst hasn't retired yet!");
+
+        core_types::RegFile rf;
```
knute-sifive, line 112: Why not initialize this variable to `core_types::RegFile::RF_INTEGER` and remove the `else` clause?

core/Rename.cpp
```diff
@@ -96,7 +104,59 @@ namespace olympia
...
+        auto dests = inst_ptr->getDestOpInfoList();
+        if(dests.size() > 0){
```
knute-sifive, line 114: Consider adding an assert in this `if` statement that there is exactly 1 destination.

core/Rename.cpp
```diff
@@ -96,7 +104,59 @@ namespace olympia
             ev_rename_insts_.schedule();
         }
     }
+    void Rename::getAckFromROB_(const InstPtr & inst_ptr)
+    {
+        sparta_assert(inst_ptr->getStatus() == Inst::Status::RETIRED,
+                        "Get ROB Ack, but the inst hasn't retired yet!");
+
+        core_types::RegFile rf;
+        auto dests = inst_ptr->getDestOpInfoList();
```
knute-sifive, line 113: This makes a copy.  Add `const &`

core/Rename.cpp
```diff
@@ -96,7 +104,59 @@ namespace olympia
...
+            rf = determineRegisterFile(dests[0]);   
+        }
+        else{
+            // in the case of no dest register, it will be a NOP
+            rf = core_types::RegFile::RF_INTEGER;
+        }
+        
+        // freeing references to PRF
+        auto dest_register_bit_mask = inst_ptr->getDestRegisterBitMask(rf);
+        auto src_register_bit_mask = inst_ptr->getSrcRegisterBitMask(rf);
```
knute-sifive, line 124: Possible copies.

core/Rename.cpp
```diff
@@ -96,7 +104,59 @@ namespace olympia
...
+        for(unsigned int i = 0; i < sparta::Scoreboard::MAX_REGISTERS; i++){
+            if(dest_register_bit_mask.test(i)){
```
knute-sifive, line 126: Consider stashing the destination register number in the instruction instead of searching for it in the mask.

core/Rename.hpp
```diff
@@ -56,13 +56,27 @@ namespace olympia
...
+        // table used for storing PRF -> ARF mapping, so we can pop the map_table entries on retirement
+        std::queue<unsigned int> reverse_map_table[2][sparta::Scoreboard::MAX_REGISTERS];
+        // mapping used when a src doesn't reference a PRF
+        // we need to track those, so that when we retire the instruction, we know not to count them
+        // towards the reference counter
+        std::queue<unsigned int> src_map_table[2][sparta::Scoreboard::MAX_REGISTERS];
+        // reference counter for PRF
+        int reference_counter[2][sparta::Scoreboard::MAX_REGISTERS] = {};
+        // list of free PRF that are available to map
+        std::queue<unsigned int> freelist[2];
```
knute-sifive, line 70: Do not use a hard-coded number (2).  This should be based on the number of register files.

knute-sifive (commented):
Great start!  More comments to follow, but some general comments:

Consider adding a small class to Instruction called `RenameData` or something to that effect that contains rename information you can use, for example, to see the assigned source and destination numbers (instead of searching through the bitmask).

test/core/rename/Rename_test.cpp
```diff
@@ -0,0 +1,324 @@
...
+        ("output_file",
+         sparta::app::named_value<std::vector<std::string>>("output_file", &datafiles),
+         "Specifies the output file")
+        ("input-file",
+         sparta::app::named_value<std::string>("INPUT_FILE", &input_file)->default_value(""),
+         "Provide a JSON instruction stream",
+         "Provide a JSON file with instructions to run through Execute")
+        ("enable_vector",
+         sparta::app::named_value<bool>("enable_vector", &enable_vector)->default_value(false),
+         "Enable the experimental vector pipelines");
```
aka2en, line 266: 😄 I believe you won't use this option

core/Rename.cpp
AaronGChan: I'm not too familiar with lambda functions, so I wasn't sure of the most elegant way to implement them. The most common example I saw were of one using a switch statement, with a case for each enum, where you would then have a specific parameter for that enum, i.e num_renames_integers:
```
switch(RegFile){
    case(RegFile::RF_INTEGER):
        populate freelist according to num_renames_rf_integers
}
```
but this seemed not too elegant, as you have to define a switch statement. I went ahead and implemented a [map](https://github.com/AaronGChan/riscv-perf-model/blob/fbe7264a97b4aadea92369bdf93a544c3077551e/core/Rename.hpp#L72) ([link](https://github.com/AaronGChan/riscv-perf-model/blob/fbe7264a97b4aadea92369bdf93a544c3077551e/core/Rename.cpp#L93) to usage in setup of Rename)  where the user can define the RegFile type and number of renames for now, but I wanted to get a better idea of the lambda approach.
Also would it be best if users define the number of renames for each register file from the .yaml file? Or would that not make sense, since you would also have to modify the core_types::RegFile enum anyways to add a RegFile?

core/Rename.cpp
AaronGChan: Removed, addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

core/Rename.cpp
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

core/Rename.cpp
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

core/Rename.cpp
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

core/Rename.cpp
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

core/Rename.cpp
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

[Link](https://github.com/AaronGChan/riscv-perf-model/blob/fbe7264a97b4aadea92369bdf93a544c3077551e/core/Inst.hpp#L35) to RenameData inner class

core/Rename.hpp
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

core/Rename.cpp
```diff
@@ -155,22 +243,30 @@ namespace olympia
...
                     const auto num = dest.field_value;
                     auto & bitmask = renaming_inst->getDestRegisterBitMask(rf);
-                    bitmask.set(num);
+                    
+                    unsigned int prf = freelist[rf].front();
+                    freelist[rf].pop();
+                    map_table[rf][num].push(prf);
+                    // reverse map table is used for mapping the PRF -> ARF
+                    // so when we get a retired instruction, we know to pop the map table entry
+                    reverse_map_table[rf][prf].push(num);
```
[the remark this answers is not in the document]
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

[Link](https://github.com/AaronGChan/riscv-perf-model/blob/fbe7264a97b4aadea92369bdf93a544c3077551e/core/Inst.hpp#L35) to RenameData inner class

core/Rename.cpp
```diff
@@ -142,10 +216,24 @@ namespace olympia
...
+                        prf = num;
+                        // we store the mapping, so when processing the retired instruction,
+                        // we know that there was no real mapping
+                        src_map_table[rf][num].push(prf);
+                    }
+                    else{
+                        prf = map_table[rf][num].back();
+                        // only increase reference counter for real references
+                        reference_counter[rf][prf]++;
+                    }
```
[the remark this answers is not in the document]
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

I store a pair that contains the src register number, and a boolean that signifies when true, that there was no prf mapped to the src register, so we don't need to decrement the reference_counter on retirement. 

[Link](https://github.com/AaronGChan/riscv-perf-model/blob/fbe7264a97b4aadea92369bdf93a544c3077551e/core/Inst.hpp#L35) to RenameData inner class

test/core/rename/Rename_test.cpp
AaronGChan: Addressed in [fbe7264](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/fbe7264a97b4aadea92369bdf93a544c3077551e)

core/Rename.cpp
```diff
@@ -113,9 +173,21 @@ namespace olympia
...
+        {
+            const auto & renaming_inst = uop_queue_.read(i);
+            const auto & dests = renaming_inst->getDestOpInfoList();
+            for(const auto & dest : dests)
+            {
+                const auto rf  = determineRegisterFile(dest);
+                rf_counters[rf]++;
+            }
+        }
+        if (credits_dispatch_ > 0 && (freelist[core_types::RegFile::RF_FLOAT].size() >= rf_counters[core_types::RegFile::RF_FLOAT] && freelist[core_types::RegFile::RF_INTEGER].size() >= rf_counters[core_types::RegFile::RF_INTEGER])) {
```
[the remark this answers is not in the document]
AaronGChan: Addressed in [6bd7115](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/6bd71153dc8a9a31dbccfacf2021cc63dcda1e45)

core/ExecutePipe.cpp
```diff
@@ -2,7 +2,7 @@
 
 #include "sparta/utils/SpartaAssert.hpp"
 #include "sparta/utils/LogUtils.hpp"
-
+#include "Rename.hpp"
```
knute-sifive, line 5: Why does ExecutePipe need to know about Rename?

core/Inst.hpp
```diff
@@ -32,6 +32,32 @@ namespace olympia
     class Inst {
     public:
 
+        class RenameData{
+            public:
```
knute-sifive, line 36: I know I haven't set up a coding style guide, but can you ensure the style you're using matches what's already in place?  For example, the `public` keyword should align with the `class` keyword:
```suggestion
        public:
```

core/Inst.hpp
```diff
@@ -32,6 +32,32 @@ namespace olympia
...
+            void setDestination(unsigned int destination){
+                dest = destination;
+            }
+            void setOriginalDestination(unsigned int destination){
+                original_dest = destination;
```
knute-sifive, line 41: Pls use `uint32_t`

core/Inst.hpp
```diff
@@ -32,6 +32,32 @@ namespace olympia
...
+            }
+            std::vector<std::pair<unsigned int, bool>> getSource(){
+                return src;
+            }
+            unsigned int getDestination(){
+                return dest;
+            }
+            unsigned int getOriginalDestination(){
+                return original_dest;
+            }
```
knute-sifive, line 54: These methods need to be `const`:
```suggestion
            const std::vector<std::pair<unsigned int, bool>> & getSource() const {
                return src;
            }
            uint32_t getDestination() const {
                return dest;
            }
            uint32_t getOriginalDestination() const {
                return original_dest;
            }
```

core/Inst.hpp
```diff
@@ -32,6 +32,32 @@ namespace olympia
...
+            private:
+            unsigned int dest;
+            unsigned int original_dest;
+            std::vector<std::pair<unsigned int, bool>> src;
```
knute-sifive, line 58: Private methods and variables end with an `_`.

core/Inst.hpp
```diff
@@ -32,6 +32,32 @@ namespace olympia
...
 
+        class RenameData{
+            public:
+            void setDestination(unsigned int destination){
+                dest = destination;
+            }
+            void setOriginalDestination(unsigned int destination){
+                original_dest = destination;
+            }
+            void setSource(std::pair<unsigned int, bool> source){
```
knute-sifive, line 43: Consider making an alias for the `pair` type: 
```
    using SourceReg = std::pair<uint32_t, bool>;
    void setSource(const SourceReg & source) {
```
And replace all references to `pair` with this alias.  That way you can change the type in 1 place (if you decide to do so).

core/Inst.hpp
```diff
@@ -163,6 +189,9 @@ namespace olympia
             return dest_reg_bit_masks_[rf];
         }
 
+        RenameData & getRenameData(){
+            return rename_data;
+        }
```
knute-sifive, line 194: Add a `const` version of this method:
```suggestion
        const RenameData & getRenameData() const {
            return rename_data;
        }
```

core/Rename.hpp
```diff
@@ -55,6 +55,24 @@ namespace olympia
...
+        std::queue<unsigned int> ** map_table = new std::queue<unsigned int>*[core_types::N_REGFILES];
+        // reference counter for PRF
+        int* reference_counter[core_types::N_REGFILES];
+        // list of free PRF that are available to map
+        std::queue<unsigned int> freelist[core_types::N_REGFILES];
+        
+        std::map<olympia::core_types::RegFile, unsigned int> NumRenameRegisters = {
+            {core_types::RegFile::RF_INTEGER, 512},
+            {core_types::RegFile::RF_FLOAT, 512}
+        };
```
knute-sifive, line 75: A few requested changes here.

1. These variables need to be in the `private` section of the class.  Need to maintain encapsulation.
2. The variable names are a mixture of different styles.  Please rename them in the style `reference_counter_` -- words separated by an `_` and end with an `_` since they are (will be) private
3. You should _never_ have "naked" pointers initialized with a raw `new`.  Even with a destructor, this is a memory leak waiting to happen (esp if an exception were thrown during construction).  _Always_ use `std::unique_ptr`.
4. The number of rename registers _per register file_ should come from parameters and not hard-coded

test/core/rename/ROBSinkUnit.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include <string>
+
+namespace rename_test
+{
+
+    ////////////////////////////////////////////////////////////////////////////////
+    // "Sink" unit, just sinks instructions sent to it.  Sends credits
+    // back as directed by params/execution mode
+    class ROBSinkUnit : public sparta::Unit
```
knute-sifive, line 20: If this class is mostly the same as `test/core/common/SinkUnit.hpp` consider deriving from it instead of copying the entire class.

core/Rename.hpp
knute-sifive: If you need public access to these variables for your tester, consider creating a RenameTester class that is a `friend` of Rename so you can access the variables as private variables.

core/Rename.cpp
knute-sifive: Sure...  this is what I had in mind:
```
    auto setup_map = [this] (core_types::RegFile reg_file, const uint32_t num_renames) {
                         for(uint32_t i = 0; i < num_renames; ++i) {
                             freelist[reg_file].emplace(i);
                         }
                         map_table[reg_file].reset(new std::queue<unsigned int>[num_renames]);
                         reference_counter[reg_file].reset(new num_renames]());
                     };

    setup_map(core_types::RegFile::RF_INTEGER, params->num_integer_renames);
    setup_map(core_types::RegFile::RF_FLOAT,   params->num_float_renames);
    static_assert(N_REGFILES == 2, "New RF type added, but Rename not updated");
```

AaronGChan:
I decided to use a pair to represent a mapping relation between ARF -> PRF, because I was reading that in typical implementation of rename with map table, they only track the youngest ARF->PRF mapping. So I replaced my queue implementation with just an array.
```
// (data value, valid bit)
using MapPair = std::pair<uint32_t, bool>;

File: test/core/rename/ROBSinkUnit.hpp (hunk 21 above)
Comment (AaronGChan) [reply test/core/rename/ROBSinkUnit.hpp]: Addressed in [163a70b](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/163a70b5862658deca0df16b035bbe59dc841c92). I had to create a separate port/function to mark instructions being sunk to retired, but that allows ROBSinkUnit to fully inherit from SinkUnit.

File: core/Rename.hpp (hunk 20 above)
Comment (AaronGChan) [reply core/Rename.hpp]: Addressed in [163a70b](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/163a70b5862658deca0df16b035bbe59dc841c92)

File: core/Inst.hpp (hunk 18 above)
Comment (AaronGChan) [reply core/Inst.hpp]: Addressed in [163a70b](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/163a70b5862658deca0df16b035bbe59dc841c92)

File: core/Inst.hpp (hunk 17 above)
Comment (AaronGChan) [reply core/Inst.hpp]: Addressed in [163a70b](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/163a70b5862658deca0df16b035bbe59dc841c92)

File: core/Inst.hpp (hunk 15 above)
Comment (AaronGChan) [reply core/Inst.hpp]: Addressed in [163a70b](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/163a70b5862658deca0df16b035bbe59dc841c92)

File: core/ExecutePipe.cpp (hunk 13 above)
Comment (AaronGChan) [reply core/ExecutePipe.cpp]: Wasn't needed, removed in [163a70b](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/163a70b5862658deca0df16b035bbe59dc841c92)

File: core/Rename.cpp (hunk 2 above)
Comment (AaronGChan) [reply core/Rename.cpp]: Thanks this made a lot more sense. Added in [163a70b](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/163a70b5862658deca0df16b035bbe59dc841c92)

File: core/Inst.hpp [hunk 22]
```diff
@@ -32,6 +32,33 @@ namespace olympia
...
+        class RenameData{
+            public:
+                void setDestination(uint32_t destination){
+                    dest_ = destination;
+                }
+                void setOriginalDestination(uint32_t destination){
+                    original_dest_ = destination;
+                }
+                using SourceReg = std::pair<uint32_t, bool>;
+                void setSource(SourceReg source){
```
knute-sifive, line 44: Pass in as `const &` to avoid the copy.

core/Inst.hpp
```diff
@@ -32,6 +32,33 @@ namespace olympia
...
+                    src_.push_back(source);
+                }
+                std::vector<std::pair<uint32_t, bool>> getSource() const {
```
knute-sifive, line 47:
```suggestion
                const std::vector<SourceReg> & getSource() const {
```

core/Inst.hpp
knute-sifive: You resolved, but didn't fix...?

core/Rename.cpp
```diff
@@ -96,7 +115,46 @@ namespace olympia
...
+        if(dests.size() > 0){
+            sparta_assert(dests.size() == 1); // we should only have one destination
+            rf = determineRegisterFile(dests[0]);   
+        }
 
+        auto const & dest = inst_ptr->getRenameData().getDestination();
+        auto const & original_dest = inst_ptr->getRenameData().getOriginalDestination();
+        auto const & srcs = inst_ptr->getRenameData().getSource();
+
+        if(dests.size() > 0){
```
knute-sifive, line 134: Duplicate `if` condition.  Can this be combined with the previous `if` statement?

core/Rename.cpp
```diff
@@ -137,15 +216,36 @@ namespace olympia
 
                 // TODO: Register renaming for sources
                 const auto & srcs = renaming_inst->getSourceOpInfoList();
+                if(srcs.size() > 1){
+                    int i = 0;
+                    i++;
+                }
```
knute-sifive, line 222: This is a weird `if` statement.  😄

knute-sifive (commented):
Looking better!

core/Inst.hpp
AaronGChan: Addressed in [3eb203c](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/3eb203c761cd11fbebea435284f6d14ba45437a7)

core/Rename.cpp
AaronGChan: Addressed in [3eb203c](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/3eb203c761cd11fbebea435284f6d14ba45437a7)

core/Rename.cpp
AaronGChan: Sorry some old debugging code 😅. Removed in [3eb203c](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/3eb203c761cd11fbebea435284f6d14ba45437a7).

klingaard:
Hey Aaron, can you merge the master branch in to your branch?  CI should be fixed.

AaronGChan:
Hey Knute,
My rename tests use the changes Zhen made in this [commit](https://github.com/sparcians/map/commit/277037f3cc2594df0ba04e3ad92d41d95e9ea3f9), which was pushed after map_v1.1.0 was released, so it doesn't contain those changes, hence the failing tests.

knute-sifive:
Ok, you can apply the attached patch to get CI working again.
[map_v1.2.0.patch](https://github.com/riscv-software-src/riscv-perf-model/files/11042670/map_v1.2.0.patch)

AaronGChan:
Hey Knute,
I was attempting to benchmark how the number of PRFs defined in the .yaml files affects the IPC to verify the renaming works, as discussed in the meeting. However, I noticed that when I drop the PRF to a low value such as 32, I get a seg fault on the freelist which stemmed from trying to pop from an empty queue. When I debugged it more, I saw that ev_rename_insts_.schedule() is able to be ran in two other lines besides the decodedInstructions_:
https://github.com/AaronGChan/riscv-perf-model/blob/41f56c239b0186de2de11a7e6490691f24ed8b63/core/Rename.cpp#L282
https://github.com/AaronGChan/riscv-perf-model/blob/41f56c239b0186de2de11a7e6490691f24ed8b63/core/Rename.cpp#L115
This is an issue because unlike the [decodedInstructions](https://github.com/AaronGChan/riscv-perf-model/blob/41f56c239b0186de2de11a7e6490691f24ed8b63/core/Rename.cpp#L163) function that has logic checking to ensure the other freelist has enough PRFs to process the instruction group, the other 2 times when  InstructionRename_ is called, there is no checking, which allowed for the freelist to be popped when it was empty, leading to the seg fault.
Does it make sense then to move the checking logic to the RenameInstructions_, and if there are not enough freelist PRFs to simply call ev_rename_insts.schedule(1) to basically sleep the rename processing until more freelist registers are allocated. With this approach, then the freelist checking doesn't have to be replicated all 3 times for when ev_rename_insts.schedule() is called.

core/Rename.hpp
```diff
@@ -109,4 +137,40 @@ namespace olympia
         ScoreboardFactories sb_facts_;
         ScoreboardTreeNodes sb_tns_;
     };
+
+    // Rename Tester Class
+    class RenameTester
```
knute-sifive, line 142: Consider moving this class to the Rename test.

knute-sifive (commented):
Awesome work.

Go ahead and add a test to `test/sim/CMakeLists.txt` that sets the renames to low values, etc to make sure it works.

core/Rename.hpp
AaronGChan: Moved to Rename_test.cpp in [128fab2](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/128fab23c955b2226bb1f60a28a32a5d96241817)

core/Rename.hpp
```diff
@@ -55,6 +58,11 @@ namespace olympia
 
         //! \brief Name of this resource. Required by sparta::UnitFactory
         static const char name[];
+        ~Rename(){
+            // delete map_table and reference counter, as they are dynamically allocated
+            map_table_.reset();
+            delete [] reference_counter_;
+        }
```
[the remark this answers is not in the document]
AaronGChan: Addressed in [128fab2](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/128fab23c955b2226bb1f60a28a32a5d96241817)

core/Inst.hpp
AaronGChan: Sorry, I missed this comment. Addressed in [d23e796](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/d23e796d1caee1fe33b8a2597d38461775489107)

knute-sifive (commented):
Looking good!  An idea for improving performance (if you want to tackle it):

When instructions enter Rename, they are in a group (`InstGroup`).  What if you categorize the instructions in the group once they are received and store it in the group?  That way, you don't have to re-calculate the resources needed every time in `scheduleRenaming`

AaronGChan:
> Looking good! An idea for improving performance (if you want to tackle it):
> 
> When instructions enter Rename, they are in a group (`InstGroup`). What if you categorize the instructions in the group once they are received and store it in the group? That way, you don't have to re-calculate the resources needed every time in `scheduleRenaming`
Are you saying to keep the instructions in the group, we get the count of the number instructions of each rf type in when we push instructions to `uop_queue_` in `decodedInstructions_`, but we keep it as a group, pass it to `scheduleRenaming` and then constantly check the`InstGroup`'s rf count? Once we have enough freelist registers to support the group, we schedule `renameInstructions_` for the entire group? Wouldn't this lead to non-optimal throughput of renamings? I.e lets say that 8 instructions can be renamed, but the group we have right now only has 6 instructions, would we just schedule a renaming only on these 6 instructions, since we're processing by group, but leave out the opportunity to rename an additional 2? 

What if instead of storing by groups, we just keep a count of the first `num_to_rename_per_cycle` instructions in the uop_queue, i.e lets say the first 8 ( assuming num_to_rename_per_cycle = 8) instructions are 6 RF_INTEGERs and 2 RF_FLOATs, we keep a global count of that in a global `rf_counters` variable. If let's say we are stalled due to lack of credit_dispatch, we keep waiting, so every time `scheduleRenaming` is called, we just compare the `rf_counters`, as that is the most number of instructions we can rename anyways. If we then get 3 `credits_dispatch` back, we can then go ahead and schedule to process 3 of the instructions. We then recategorize the first 8 instructions again, get update the rf_counters, and then continue on. This approach wouldn't be blocking, because we don't have to wait for a group, and if we can rename for the current group, as well as the next group, we're able to because we're tracking the top X amount of instructions. We also wouldn't need to pass around a `InstGroup` around as well.

Not sure if I misinterpreted your idea, so please correct me if I got it wrong.

knute-sifive:
So my suggestion was stemmed from the thinking that Rename would *only* rename (or make forward progress) iff the _entire_ group of instructions could be renamed in 1 cycle.  If you'd like to support _partial_ renaming, then you're right, Rename should be more flexible.

To handle both use cases (partial or all), as instructions enter Rename from Decode, you could categorize what is in the the `uop_queue` for every index.  Some sloppy pseudo code:

```
struct RegCountData {
    uint32_t cumulative_reg_counts[core_types::RegFile::N_REGFILES] = {0};
};
RegCountData uop_queue_regcount_data_[uop_queue_size];
```
and as you move an instruction from the `InstGroup` into the `uop_queue`:
```
    uint32_t uop_idx = 0;
    RegCountData current_counts; // starts at 0
    for (inst in group) {
         uop_queue_.insert(inst);
         current_counts = accumulateCounts_(current_counts, getRegCountData(inst));
         uop_queue_regcount_data_[uop_idx] = current_counts;
    }
```
then, when you attempt renaming (in `scheduleRenaming_`) you can see how far you can rename:
```
    // Start at the end of the queue.  You hope you can rename all of them.  
    // But it might make more sense to start at the beginning...
    uint32_t num_to_rename = uop_queue_.size();
    for(counts in uop_queue_regcount_data_,rbegin()) {
         if(counts < current_rename_resources) {
             break;
         }
         --num_to_rename;
    }
    if(num_to_rename == 0) {
        // no resources, go back to sleep.
    }
```
Anyway, just some thoughts.  If you want to tackle this in another PR, that's ok by me.

AaronGChan:
> So my suggestion was stemmed from the thinking that Rename would _only_ rename (or make forward progress) iff the _entire_ group of instructions could be renamed in 1 cycle. If you'd like to support _partial_ renaming, then you're right, Rename should be more flexible.
> 
> To handle both use cases (partial or all), as instructions enter Rename from Decode, you could categorize what is in the `uop_queue` for every index. Some sloppy pseudo code:
> 
> ```
> struct RegCountData {
>     uint32_t cumulative_reg_counts[core_types::RegFile::N_REGFILES] = {0};
> };
> RegCountData uop_queue_regcount_data_[uop_queue_size];
> ```
> 
> and as you move an instruction from the `InstGroup` into the `uop_queue`:
> 
> ```
>     uint32_t uop_idx = 0;
>     RegCountData current_counts; // starts at 0
>     for (inst in group) {
>          uop_queue_.insert(inst);
>          current_counts = accumulateCounts_(current_counts, getRegCountData(inst));
>          uop_queue_regcount_data_[uop_idx] = current_counts;
>     }
> ```
> 
> then, when you attempt renaming (in `scheduleRenaming_`) you can see how far you can rename:
> 
> ```
>     // Start at the end of the queue.  You hope you can rename all of them.  
>     // But it might make more sense to start at the beginning...
>     uint32_t num_to_rename = uop_queue_.size();
>     for(counts in uop_queue_regcount_data_.rbegin()) {
>          if(counts < current_rename_resources) {
>              break;
>          }
>          --num_to_rename;
>     }
>     if(num_to_rename == 0) {
>         // no resources, go back to sleep.
>     }
> ```
> 
> Anyway, just some thoughts. If you want to tackle this in another PR, that's ok by me.

Implemented partial renaming in [8e2452c](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/8e2452c60dfd7ee3648501b5ca1b926a4961d6bd)

core/Rename.cpp
```diff
@@ -142,10 +258,27 @@ namespace olympia
                     const auto rf  = determineRegisterFile(src);
                     const auto num = src.field_value;
                     auto & bitmask = renaming_inst->getSrcRegisterBitMask(rf);
-                    bitmask.set(num);
+                    uint32_t prf;
```
knute-sifive, line 261: Always initialize your variables to some invalid value -- in case someone modifies your code and forgets to provide a final value.  (most compilers catch that, but not always...)

core/Rename.cpp
```diff
@@ -93,10 +114,49 @@ namespace olympia
...
+                freelist_[rf].push(dest);
+            }
+            map_table_[rf][original_dest].first = 0;
+            // map entry is now invalid, new srcs should read from ARF
+            map_table_[rf][original_dest].second = false;
+        }
 
+        auto const & srcs = inst_ptr->getRenameData().getSource();
+        // freeing references to PRF
+        for(auto src: srcs){
```
knute-sifive, line 143: Consider `const auto &` to avoid copy

core/Rename.cpp
```diff
@@ -40,7 +40,28 @@ namespace olympia
             registerConsumerHandler(CREATE_SPARTA_HANDLER_WITH_DATA(Rename, creditsDispatchQueue_, uint32_t));
         in_reorder_flush_.
             registerConsumerHandler(CREATE_SPARTA_HANDLER_WITH_DATA(Rename, handleFlush_, FlushManager::FlushingCriteria));
+        in_rename_retire_ack_.registerConsumerHandler
+                (CREATE_SPARTA_HANDLER_WITH_DATA(Rename, getAckFromROB_, InstPtr));
         sparta::StartupEvent(node, CREATE_SPARTA_HANDLER(Rename, setupRename_));
+        
+        auto setup_map = [this] (core_types::RegFile reg_file, const uint32_t num_renames) {
+                        for(uint32_t i = 0; i < num_renames; ++i) {
```
knute-sifive, line 48: I _think_ this is a bug.  On line 97 below, you initialize the first 32 registers as already ready (0->31).  But here the free renames _start_ at 0 and will _include_ those renames already set as ready.  In essence they are already in use (and not free).  I think this is what you want:

```suggestion
                        sparta_assert(num_regs < num_renames);
                        for(uint32_t i = num_regs; i < num_renames; ++i) {
```

knute-sifive:
One more thing to add and I think this is good to go: can you add stall counters?  Specifically, you know when you performed no renaming compared to a partial renaming compared to a full renaming.  Maybe add a histogram of how many instructions were renamed (0, 1, 2, ... max).

core/Rename.cpp
AaronGChan: If we count the marking of registers as ready at line 97 and never add them to the freelist at the beginning, then we will never access those "ready" registers, because we only push back into the freelist register destinations that are retired. 
Are we treating those registers marked as ready at line 97 as the initial ARFs, and then as we progress through renaming, the ARFs get mapped to a new PRF? We would then need a way to track what is the current "valid" ARF value and then when a current instruction with a destination to a register is finished, we can then retire the "valid" ARF register for that one.

knute-sifive: I don't think there's need to track ARFs separately...

I might have missed this in your implementation -- do you already have a mapping for those first 32 registers?  If not, then this might the difference in our thinking.  I have a suggestion for fix below.

Consider this example as the very first instruction in the trace:
```
     add x1, x2, x3
```
Now, in a real machine, if this were the first instruction, you're technically already borked -- what are the values for `x2` and `x3`?  They've never been written to!  In some implementations, these registers are mapped to either a null register or x0.  But in the perf model, you've already accounted for these registers by setting `x0 -> x31` as ready on line 97 (implying they've been written to).   Your register map should have that mapping reflected as well:
```
    for(uint32_t i = 0; i < num_regs; ++i) {
        map_table_[reg_file][i] = {i, false};
    }
```

After renaming (and my suggested fixes):
```
    add x1, x2, x3  -> add x1 -> x32, x2 -> x2, x3 -> x3
```
When the `add` instruction retires, `x1` will be placed onto the freelist.

But you make an interesting observation -- if I ask for 48 renames, does that include the original 32?  Technically, if the map were setup so that x0 -> x0 (not real, BTW), x1 -> x1, etc, then you actually have 48 - 32 or 16 actual "free" renames for use.

arupc:
Hi Aaron and Knute, I added a comment in the rename discussion thread, about creating a rename API. Please review and let me know what you think.

core/Inst.hpp
```diff
@@ -32,6 +32,34 @@ namespace olympia
     class Inst {
     public:
 
+        class RenameData{
+            public:
+                using SourceReg = std::pair<uint32_t, bool>;
+                
+                void setDestination(uint32_t destination){
```
aka2en, line 39: Consider using these namings to help understand:
setDestPhysReg
setDestArchReg
setArchPhysRegs
setArchPhysRegs

core/Rename.cpp
```diff
@@ -142,10 +258,27 @@ namespace olympia
                     const auto rf  = determineRegisterFile(src);
                     const auto num = src.field_value;
                     auto & bitmask = renaming_inst->getSrcRegisterBitMask(rf);
-                    bitmask.set(num);
+                    uint32_t prf;
+
+                    bool no_prf = false;
+                    if(!map_table_[rf][num].second){
+                        prf = num;
```
aka2en, line 265: Does this line indicate PRF#1, PRF#2, ... PRF#31 are reserved for ARF x1, x2, ... x31?

core/Rename.cpp
```diff
@@ -142,10 +258,27 @@ namespace olympia
...
+                    }
+                    else{
+                        prf = map_table_[rf][num].first;
+                        // only increase reference counter for real references
+                        reference_counter_[rf][prf]++;
+                    }
+                    olympia::Inst::RenameData::SourceReg rename_data_src(prf, no_prf);
+                    renaming_inst->getRenameData().setSource(rename_data_src);
+                    bitmask.set(prf);
+                    scoreboards_[rf]->set(bitmask);
```
aka2en, line 278: Wanna make sure if it's your desired behavior: this would make source PRFs ready in Rename. Which means, operand readiness are set by itself, not other dependent instructions.

core/Rename.cpp
```diff
@@ -155,28 +288,36 @@ namespace olympia
...
+
+                    uint32_t prf = freelist_[rf].front();
+                    freelist_[rf].pop();
+                    map_table_[rf][num].first = prf;
+                    map_table_[rf][num].second = true;
+
+                    renaming_inst->getRenameData().setDestination(prf);
+                    renaming_inst->getRenameData().setOriginalDestination(num);
+                    bitmask.set(prf);
+                    scoreboards_[rf]->set(bitmask);
```
aka2en, line 300: Same comment

aka2en (commented):
I like how it is shaped -- surprised it's done by 1 man :) It would be great to have a document (and figures) describing how register freelist works. By that we could examine the idea and codes efficiently.

AaronGChan:
> I like how it is shaped -- surprised it's done by 1 man :) It would be great to have a document (and figures) describing how register freelist works. By that we could examine the idea and codes efficiently.

Added [documentation](https://github.com/riscv-software-src/riscv-perf-model/discussions/19#discussioncomment-5626736) for current rename implementations

core/Rename.cpp
AaronGChan: Implemented in [1bf68f4](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/1bf68f4f79753f81792e51ef39c2b06e261e82de)

core/Rename.cpp
AaronGChan: Addressed in [1bf68f4](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/1bf68f4f79753f81792e51ef39c2b06e261e82de)

core/Rename.cpp
AaronGChan: Addressed in [1bf68f4](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/1bf68f4f79753f81792e51ef39c2b06e261e82de)

AaronGChan:
> One more thing to add and I think this is good to go: can you add stall counters? Specifically, you know when you performed no renaming compared to a partial renaming compared to a full renaming. Maybe add a histogram of how many instructions were renamed (0, 1, 2, ... max).

Implemented in [1bf68f4](https://github.com/riscv-software-src/riscv-perf-model/pull/40/commits/1bf68f4f79753f81792e51ef39c2b06e261e82de)

core/Rename.cpp
AaronGChan: For this PR, I'm only implementing the renaming structures, but I'm not checking for operand dependencies yet, that'll be the next task. So I'm just marking all instructions as ready to allow them to execute.

core/Rename.cpp
AaronGChan: Sorry this was older code, Knute and I have discussed that the ARF register should be mapped to the PRF (x0 - x31 mapped to PRF0 - PRF31), so this code isn't applicable anymore. When I wrote this i was under the assumption that once an instruction retired and no references and no mappings, then we would point to the ARF instead, but we're now assuming there is always a valid ARF->PRF mapping.

core/Inst.hpp
AaronGChan: Since we're keeping all references in the PRF, since we initialized the ARF into the PRF at startup, I think the current naming of "original" destination and destination make more sense. We want to know at retirement what was the "original" or older destination before we defined a new mapping, because we replace the map_table_ mapping when we map a new destination. We then can retire the original destination if it doesn't have any references, or we retire it when the last reference to it is retired.

AaronGChan:
Is a reference_counter actually needed? We're already initializing PRFs at the beginning (x1 -> PRF1, and so on), so when a new mapping is allocated, lets say x1 -> PRF32, then when that instruction is retired x1 is now officially mapped to PRF32, any source operands for any instructions issued after the new mapping instruction will now point to PRF32, because it would reference the new map_table. Essentially, when a new rename mapping is officially retired, we know that the old rename PRF is now free, without having to use a reference_counter.

Example:
I1: ADD x1, x2, x3
I2: ADD x2, x1, x3
I3: ADD x1, x7, x9
I4: ADD x5, x1, x11

Current Mappings:
x1 -> PRF1
x2 -> PRF2
...
x31 -> PRF31

Rename I1 (ADD x1, x2, x3)
x1 -> PRF32
map_table is set for PRF32, so x1 now points to PRF32, so we read from PRF32

New Mappings:
x1 -> PRF32
.. everything else is the same

Rename I2 (ADD x2, x1, x3)
x2 -> PRF33
Reads PRF32 for x2

Rename I3 (ADD x1, x7, x9)
x1 -> PRF34

New Mappings:
x1 -> PRF34
.. everything else is the same

Rename I4 (ADD x5, x1, x11)
x5 -> PRF35

map_table is set for PRF34, so x1 now points to PRF32, so we read from PRF34


And then lets say we then start retiring:
Retire I1 (ADD x1, x2, x3)
We know PRF1 is now free, because I1 is now retiring, no instructions were issued before this one.
Retire I2 (ADD x2, x1, x3)
We know PRF2 is now free, because l2 is now retiring.
Retire I3 (ADD x1, x7, x9)
We know PRF32 is now free, because l3 is now retiring. Any instruction after l3 would have used the new mapping of x1 -> PRF34. In the case of instruction I4, it read from PRF34.

knute-sifive:
> Is a reference_counter actually needed?

In the current implementation, you are correct, it's not needed. But it's a nice to have for optimizations like [move elimination](https://easyperf.net/blog/2018/04/22/What-optimizations-you-can-expect-from-CPU) or other ... stuff.  😁 

Since you're in the middle of adding support for Renames, might as well think ahead. 😉

knute-sifive (approved):
This is looking good.  Let's make some progress on this.  We can make more optimizations later.  Right now, please start on the next phase!

AaronGChan:
I'm trying to build the conda environment to test my code on new changes, and I'm running into this error:
I'm using an M1 macbook air with macOS Ventura 13.3.1
```
(base) aaronchan@Aarons-MacBook-Air-2 conda % conda env create -f environment.yml                                

Collecting package metadata (repodata.json): done
Solving environment: failed

ResolvePackageNotFound: 
  - libnsl[version='>=2.0.0',build=h7f98852_0]
  - libblas[version='>=3.9.0',build=16_linux64_openblas]
  - expat[version='>=2.5.0',build=h27087fc_0]
  - pcre[version='>=8.45',build=h9c3ff4c_0]
  - libstdcxx-ng[version='>=12.2.0',build=h46fd767_19]
  - zlib[version='>=1.2.13',build=h166bdaf_4]
  - sqlite[version='>=3.36.0',build=h9cd32fc_2]
  - yaml-cpp[version='>=0.7.0',build=h27087fc_2]
  - xz[version='>=5.2.6',build=h166bdaf_0]
  - c-ares[version='>=1.18.1',build=h7f98852_0]
  - libstdcxx-devel_linux-64[version='>=10.4.0',build=hd38fd1e_19]
  - libiconv[version='>=1.17',build=h166bdaf_0]
  - libzlib[version='>=1.2.13',build=h166bdaf_4]
  - ld_impl_linux-64[version='>=2.39',build=hcc3a1bd_1]
  - openssl[version='>=1.1.1s',build=h0b41bf4_1]
  - libgfortran-ng[version='>=12.2.0',build=h69a702a_19]
  - libssh2[version='>=1.10.0',build=haa6b8db_3]
  - readline[version='>=8.1.2',build=h0f457ee_0]
  - gcc_linux-64[version='>=10.4.0',build=h9215b83_11]
  - libcblas[version='>=3.9.0',build=16_linux64_openblas]
  - libgcc-ng[version='>=12.2.0',build=h65d4601_19]
  - liblapack[version='>=3.9.0',build=16_linux64_openblas]
  - binutils_impl_linux-64[version='>=2.39',build=he00db2b_1]
  - hdf5[version='>=1.10.6',build=nompi_h6a2412b_1114]
  - numpy[version='>=1.24.1',build=py310h08bbf29_0]
  - libuv[version='>=1.44.2',build=h166bdaf_0]
  - gxx_impl_linux-64[version='>=10.4.0',build=h5231bdf_19]
  - libffi[version='>=3.4.2',build=h7f98852_5]
  - libsqlite[version='>=3.40.0',build=h753d276_0]
  - libcurl[version='>=7.87.0',build=h6312ad2_0]
  - doxygen[version='>=1.8.20',build=had0d8f1_0]
  - libgcc-devel_linux-64[version='>=10.4.0',build=hd38fd1e_19]
  - gcc_impl_linux-64[version='>=10.4.0',build=h5231bdf_19]
  - keyutils[version='>=1.6.1',build=h166bdaf_0]
  - ca-certificates[version='>=2022.12.7',build=ha878542_0]
  - binutils_linux-64[version='>=2.39',build=h5fc0e48_11]
  - curl[version='>=7.87.0',build=h6312ad2_0]
  - ncurses[version='>=6.3',build=h27087fc_1]
  - boost-cpp[version='>=1.76.0',build=h312852a_1]
  - python[version='>=3.10.8',build=h257c98d_0_cpython]
  - libsanitizer[version='>=10.4.0',build=h5246dfb_19]
  - libgfortran5[version='>=12.2.0',build=h337968e_19]
  - rapidjson[version='>=1.1.0',build=he1b5a44_1002]
  - icu[version='>=68.2',build=h9c3ff4c_0]
  - rhash[version='>=1.4.3',build=h166bdaf_0]
  - _libgcc_mutex[version='>=0.1',build=conda_forge]
  - _openmp_mutex[version='>=4.5',build=2_gnu]
  - gxx_linux-64[version='>=10.4.0',build=h6e491c6_11]
  - libgomp[version='>=12.2.0',build=h65d4601_19]
  - libnghttp2[version='>=1.51.0',build=hdcd2b5c_0]
  - bzip2[version='>=1.0.8',build=h7f98852_4]
  - libedit[version='>=3.1.20191231',build=he28a2e2_2]
  - tk[version='>=8.6.12',build=h27826a3_0]
  - libuuid[version='>=2.32.1',build=h7f98852_1000]
  - boost[version='>=1.76.0',build=py310h7c3ba0c_1]
  - libev[version='>=4.33',build=h516909a_1]
  - zstd[version='>=1.5.2',build=h6239696_4]
  - libopenblas[version='>=0.3.21',build=pthreads_h78a6416_3]
  - cppcheck[version='>=2.7.5',build=py310h94ea96f_1]
  - krb5[version='>=1.20.1',build=hf9c8cef_0]
```

AaronGChan:
Spent a lot of time on the errors around this issue, and I didn't realize there was an issue open for it until after I figured out the same fix:
https://github.com/sparcians/map/issues/418
I think it would be helpful to maybe link this issue in the Readme when trying to build Olympia, because I got this issue both in Sparta and in Olympia when trying to build when trying to build on my m1 machine

AaronGChan:
This should be ready to merge? Or should I merge the rest of my operand dependency changes into this PR to have all of the operand dependency changes under one PR?

knute-sifive:
I think you should merge this one in and start a new PR with the next phase.

AaronGChan:
I don't have write access to merge PRs, could you merge it?

PR #42, opened 2023-03-13
Update boost deps for core and mss, track Sparta PR397

### LINKED ISSUE #35: Boost 1.81 not working (on macOS) - non-default boost location not supported by CMakeLists

peter-d:
This fixes #35 and tracks the changes in Sparta [PR397](https://github.com/sparcians/map/pull/397)


knute-sifive:
Ok, step 1 complete: MAP PR merged: https://github.com/sparcians/map/pull/397

peter-d:
@knute-sifive I'm kind of stuck at this point: I have the HDF5 issue fixed, but with the CI pinned to sparta v1.1.1 this won't be able to pass...

Any advice? I don't want to break master CI either...

knute-sifive:
Sorry, been trying to find some time to get back to this!  I can help if you get stuck.

knute-sifive:
Ah, just caught up.  I'll take a look

knute-sifive:
Seeing if I can reproduce locally...

knute-sifive:
The issue on the release build is related to LTO.  Looking into the magic to make it work...

knute-sifive:
Hazaa!

knute-sifive:
Hey Peter, update your clone and see if a release and/or debug build works for you.

core/CMakeLists.txt
```diff
@@ -20,5 +20,3 @@ add_library(core
   CPUFactory.cpp
   CPUTopology.cpp
   )
-target_include_directories(core SYSTEM PRIVATE ${Boost_INCLUDE_DIRS})
-target_link_libraries(core PUBLIC SPARTA::sparta)
```
peter-d: Since I don't do a conda install of sparta, this fails for me: sparta  include paths and link paths are not passed properly.

peter-d:
I had to put back a few target_link_libraries to find sparta (I don't have it installed in conda env so it's not picked up automatically in the default paths).

Since that does not break CI, this looks good to go to me, but it could do with a review especially to check:
- is it ok to pull in mavis and stf submodules from current masters? I updated the references
- the environment.yml file was coming from a `conda list` but I don't run in conda, so I manually edited the file. It currently works, but maybe is not what we want.

knute-sifive:
> I had to put back a few target_link_libraries to find sparta

That's so weird that's required for the core library build.  Curious -- what didn't link?  The issue I have with adding sparta as a link library is that I believe this causes the core archive to blow up in size.  I think.

> is it ok to pull in mavis and stf submodules from current masters?

I don't see problem with that.  Probably should use an stf tag and tag Mavis so we have consistency and not some random SHA.  I'll look into that on the next revision.

Updating the yaml file for Conda is fine -- the CI uses it, so if it passes, then it should good to go!

knute-sifive:
> I had to put back a few target_link_libraries to find sparta

That's so weird that's required for the core library build.  Curious -- what didn't link?  The issue I have with adding sparta as a link library is that I believe this causes the core archive to blow up in size.  I think.

> is it ok to pull in mavis and stf submodules from current masters?

I don't see problem with that.  Probably should use an stf tag and tag Mavis so we have consistency and not some random SHA.  I'll look into that on the next revision.

Updating the yaml file for Conda is fine -- the CI uses it, so if it passes, then it should good to go!

knute-sifive:
> I had to put back a few target_link_libraries to find sparta

That's so weird that's required for the core library build.  Curious -- what didn't link?  The issue I have with adding sparta as a link library is that I believe this causes the core archive to blow up in size.  I think.

> is it ok to pull in mavis and stf submodules from current masters?

I don't see problem with that.  Probably should use an stf tag and tag Mavis so we have consistency and not some random SHA.  I'll look into that on the next revision.

Updating the yaml file for Conda is fine -- the CI uses it, so if it passes, then it should good to go!

peter-d:
I added `target_link_libraries(... SPARTA::sparta)` since that then automatically also sets the right include paths for sparta when compiling. But I did not really consider the impact it might have at linking stage...

Since it's all statically linked at the end, I guess it could be sufficient to just add the includes here?

CMakeLists.txt
```diff
@@ -11,16 +11,37 @@ add_compile_options ( -Werror
...
+set(CMAKE_MODULE_PATH "${SPARTA_SEARCH_DIR}/lib/cmake/sparta" ${CMAKE_MODULE_PATH})
+
+find_package(Sparta REQUIRED)
+
+if (NOT SPARTA_FOUND)
+  message (FATAL_ERROR "Sparta was not found in ${SPARTA_SEARCH_DIR}")
+else()
+  message (STATUS "Sparta was found in ${SPARTA_SEARCH_DIR}")
+endif()
+#include_directories(${SPARTA_INCLUDE_DIRS})
```
knute-sifive, line 31: I'm wondering if uncommenting this line with fix your include issues.  Can you do that AND remove the target_link lines to see if you still have problems?

peter-d: It's what this line and the next in `FindSparta.cmake` are providing automatically when you add sparta as a link library:

https://github.com/sparcians/map/blob/6f0636f7a717238cd039be33f7975048449bb82c/sparta/cmake/FindSparta.cmake#L69

knute-sifive: Ok, I can reproduce the missing include problem, but the not boost one.  I'll admit that I don't fully grok how `cmake` brings in the goods from a package.  You'd think the `find_package` method would be enough!  But sparta must be installed in a "standard" location or something so that includes AND library paths are set up?  I don't get it...

peter-d: All `find_package` does is import a target for the library (as if it were defined elsewhere in the makefiles with `add_library`) but it won’t set up any default compile or link flags project wide, it will only add the relevant options to the targets when you explicitly say they depend on the imported target. In this case that means unless you explicitly add SPARTA::sparta it won’t pass the flags. Which you won’t notice if everything is installed in system wide include and lib paths, but it will break as soon as one of the components (either Sparta or the lobs it depends on) is not in that default path. 

That said, there could be a way to only add the include paths and maybe not the link flags, but I expect that to fail. 

Reason I see boost problems is because I have multiple boost versions via homebrew, and Sparta needs one that is not the default one. (There’s macOS issues with boost 1.81 command line arguments)

arupc: Thank you Peter and Knute for working through this. 

Trying to understand the latest status of this work: From the comments above, I understand at least the regression is not working yet. But is it just the regression or more than that? Would I be able to build riscv-perf-model / olympia with latest head of master of Sparta (which installs Sparta in the conda environment)?

peter-d: As long as Sparta (and it’s dependencies like boost hdf5 etc) are available in the default paths (line and includes) it just works. It’s just in case some dependencies are not in a default location , that the overridden paths are not passed through from Sparta to Olympia at this point… (eg I have a home brew boost version in macOS that is not the default one which gives me troubles when I want to this locally branch currently)

peter-d:
@knute-sifive if you could review? 
For me this is good to go (if debug CI won't fail for a weird reason)
Thanks!

knute-sifive:
Absolutely!  Yes, let's get this in by the weekend.

.github/workflows/ubuntu-build.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         with:
           repository: sparcians/map
           path: map
-          ref: map_v1.1.1
+          ref: map_v2.0.0
```
peter-d, line 39: Is this ready to merge with the PR in sparta now merged as well? This checks out map_v2.0.0 tag of sparta, but I guess that can stay?

knute-sifive: I'd like to make a release of Sparta with that change OR we can just use `map_v2` development branch (will cherry pick your merge).

klingaard: Pushed a change to map_v2.  Once regression passes, I'll merge this in!

knute-sifive (approved):

PR #45, opened 2023-03-20
Clone specific version of map for CI

knute-sifive:


knute-sifive:
@arupc can you approve this?

arupc (approved):

PR #46, opened 2023-04-18
ci: Use all cpu cores

h0lyalg0rithm:
Small to change to  use all the CPU cores for the build.


knute-sifive (approved):
Appreciate the change!

PR #51, opened 2023-05-19
Issue #2: Need Operand Dependency Checking Task 2

AaronGChan:
Summary of Changes:
LSU.cpp -> Operand dependency checking based on source bits in scoreboard, callbacks
ExecutePipe.cpp -> Operand dependency checking based on source bits in scoreboard, callbacks
Rename_test.cpp -> Added testing for both LSU and ExecutePipe with RAW instruction jsons, imported OlympiaSim for full simulation run to access data

Final portion of operand dependency checking implemented in this PR for LSU and ExecutePipe.


core/ExecutePipe.cpp
```diff
@@ -69,35 +69,43 @@ namespace olympia
     {
         // FIXME: Now every source operand should be ready
         const auto & src_bits = ex_inst->getSrcRegisterBitMask(reg_file_);
-        sparta_assert(scoreboard_views_[reg_file_]->isSet(src_bits),
+        if(scoreboard_views_[reg_file_]->isSet(src_bits)){
+            sparta_assert(scoreboard_views_[reg_file_]->isSet(src_bits),
                       "Should be all ready source operands ... " << ex_inst);
```
klingaard, line 74: Probably can remove this assert.

core/ExecutePipe.hpp
```diff
@@ -124,6 +124,9 @@ namespace olympia
 
         // Used to flush the ALU
         void flushInst_(const FlushManager::FlushingCriteria & criteria);
+
+        // Friend class used in rename testing
+        friend class ExecutePipeTester;
```
klingaard, line 129: Seems this class is missing from the PR

core/LSU.cpp
```diff
@@ -1,9 +1,14 @@
...
 #include "sparta/utils/SpartaAssert.hpp"
 #include "LSU.hpp"
 
 namespace olympia
 {
+    inline core_types::RegFile determineRegisterFile(const mavis::OperandInfo::Element & reg)
+    {
+        const bool is_float = (reg.operand_type == mavis::InstMetaData::OperandTypes::SINGLE) ||
+                              (reg.operand_type == mavis::InstMetaData::OperandTypes::DOUBLE);
+        return is_float ? core_types::RegFile::RF_FLOAT : core_types::RegFile::RF_INTEGER;
```
klingaard, line 10: This is great, but to flag an early error to future developers, consider a full check.  And maybe create a map for speed:
```suggestion
        static const std::map<mavis::InstMetaData::OperandTypes, core_types::RegFile> mavis_optype_to_regfile = { 
            // mapping of supported types ... 
         };

        if(auto match = mavis_optype_to_regfile.find(reg); match != mavis_optype_to_regfile.end()) {
            return match->second;
        }
        sparta_assert(false, "Unknown reg type: " << ... more info);
        return core_types::RegFile::UNKNOWN;

```

core/LSU.cpp
```diff
@@ -1,9 +1,14 @@
-
 #include "sparta/utils/SpartaAssert.hpp"
 #include "LSU.hpp"
 
 namespace olympia
 {
+    inline core_types::RegFile determineRegisterFile(const mavis::OperandInfo::Element & reg)
```
klingaard, line 6: Can you move this handy method to CoreUtils.hpp?

core/LSU.cpp
```diff
@@ -87,48 +91,78 @@ namespace olympia
...
-        updateIssuePriorityAfterNewDispatch_(inst_ptr);
-        uev_issue_inst_.schedule(sparta::Clock::Cycle(0));
+    {   
+        core_types::RegFile reg_file = core_types::RF_INTEGER;
+        const auto & srcs = inst_ptr->getSourceOpInfoList();
+        const auto & dests = inst_ptr->getDestOpInfoList();
+        if(dests.size() > 0){
+            sparta_assert(dests.size() == 1); // we should only have one destination
+            reg_file = determineRegisterFile(dests[0]);
+        }
```
klingaard, line 116: Why are you using the destination register as the register file type?

test/core/rename/Rename_test.cpp
```diff
@@ -341,25 +372,70 @@ void runTest(int argc, char **argv)
...
+        uint32_t num_cores = 1;
+        bool show_factories = false;
+        OlympiaSim sim("simple",
+                       scheduler,
+                       num_cores, // cores
+                       input_file,
+                       ilimit,
+                       show_factories);
+        cls.populateSimulation(&sim);
+        sparta::RootTreeNode* root_node = sim.getRoot();
```
klingaard, line 408: This code is the same regardless of the input_file.  Can probably be moved outside of the `if-else`

core/LSU.cpp
AaronGChan: Addressed in [4fd928d](https://github.com/riscv-software-src/riscv-perf-model/pull/51/commits/4fd928d4400af49de00ac3f5f8b5c4990bdbb2fd)

core/ExecutePipe.cpp
AaronGChan: Addressed in [4fd928d](https://github.com/riscv-software-src/riscv-perf-model/pull/51/commits/4fd928d4400af49de00ac3f5f8b5c4990bdbb2fd)

core/ExecutePipe.hpp
AaronGChan: Defined in [Rename_Test](https://github.com/riscv-software-src/riscv-perf-model/blob/4fd928d4400af49de00ac3f5f8b5c4990bdbb2fd/test/core/rename/Rename_test.cpp#L102)

core/LSU.cpp
AaronGChan: Addressed in [4fd928d](https://github.com/riscv-software-src/riscv-perf-model/pull/51/commits/4fd928d4400af49de00ac3f5f8b5c4990bdbb2fd)

test/core/rename/Rename_test.cpp
AaronGChan: Addressed in [4fd928d](https://github.com/riscv-software-src/riscv-perf-model/pull/51/commits/4fd928d4400af49de00ac3f5f8b5c4990bdbb2fd)

core/LSU.cpp
AaronGChan: I used the destination register for the register file type because we have to mark the destination register as ready for loads in the LSU in the corresponding scoreboard.

core/LSU.cpp
```diff
@@ -87,48 +86,76 @@ namespace olympia
...
+        core_types::RegFile reg_file = core_types::RF_INTEGER;
+        const auto & srcs = inst_ptr->getSourceOpInfoList();
+        const auto & dests = inst_ptr->getDestOpInfoList();
+        if(dests.size() > 0){
+            sparta_assert(dests.size() == 1); // we should only have one destination
+            reg_file = olympia::coreutils::determineRegisterFile(dests[0]);
+        }
+        else if(srcs.size() > 0){
+            reg_file = olympia::coreutils::determineRegisterFile(srcs[0]);
+        }
```
knute-sifive, line 114: Sorry, got to poke at this again...

You should only be looking at the instruction's sources for readiness (`isSet`) not its destination.  You are correct that you use the destination register file type to set the scoreboard, but this code doesn't do that here...

knute-sifive:
Another general comment:

Store operations actually have 2 source types: one for address calculation (integer) and one for store data (integer, float, vector).  We will eventually need to distinguish between the two in the LS.  The address sources are the most important -- need to start the process of MMU, cache lookup, etc.  But the store data... that can technically wait.

core/LSU.cpp
AaronGChan: Addressed in [6c945e3](https://github.com/riscv-software-src/riscv-perf-model/pull/51/commits/6c945e37863a66ec5402acd2cb98f08f006c9221)

knute-sifive:
I think this is ready. What do you think?

knute-sifive:
> > I think this is ready. What do you think?
> 
> Did you want me to incorporate changes to address for stores if the address calculation is ready but the store data isn't? Or did you want to open a separate issue as an enhancement to address it?

Didn't look at the code in great detail, but what does the LS do now?  Does it wait for all operands to be ready before issuing the store?  Or does it ignore part of it?

knute-sifive (approved):

AaronGChan:
Could you merge the PR? I don't have write access.

PR #59, opened 2023-06-29
i2f Bug Fix

AaronGChan:
Changing ExecutePipe.cpp code to set scoreboard based on destination register, not the pipe register. 
Added associated test for checking correct register file's rename registers are used.


knute-sifive (approved):
Great fix!  And I like the testing to go with it.  ;)

Issue #58, opened 2023-06-29
Transfer instructions (F2I, I2F) update wrong scoreboard


knute-sifive:
The following sequence of instructions cause olympia to timeout:
```
[
    {
        "mnemonic": "fmv.d.x",
        "rs1": 1,
        "fd": 17
    },
    {
        "mnemonic": "fadd.d",
        "fs1": 10,
        "fs2": 17,
        "fd": 7
    }
]
```
The problem is that the `fmv` instruction is dispatched to the IEX block to read `x1`, but when that pipeline writes the `f17` rename, it's actually writing to a random rename for `x17`.

The code that needs modification: https://github.com/riscv-software-src/riscv-perf-model/blob/f7c17dcb8b10f80650fa50cbc1affe31e3a0624f/core/ExecutePipe.cpp#L142C2-L143

The code should not assume the destination register file is the same as the unit's.

knute-sifive:
Fixed.

PR #63, opened 2023-06-30
Fixed another issue related to i2f; added pipe support/debug messages

knute-sifive:
Couple more things:
- Improved the performance of the simulator by reducing retire events
- Added param support to enable/disable skipping of non-user mode instructions in an STF
- Added more debug messages
- Added Rename stall conditions/counters
- Fixed instruction status; was not being updated


knute-sifive:
closes #58

AaronGChan (approved):
Looks good to me!

PR #66, opened 2023-07-11
Add documentation for coremark

light2802:
This patchset adds documentation for generating coremark trace from scratch.


traces/stf_trace_gen/coremark_riscv.patch
```diff
@@ -0,0 +1,835 @@
...
+ #endif
++	STOP_TRACE
+     stop_time();
+     total_time = get_time();
+     /* get a function of the input to report */
+diff --git a/riscv64/core_portme.c b/riscv64/core_portme.c
+new file mode 100755
+index 0000000..8f17cb8
+--- /dev/null
++++ b/riscv64/core_portme.c
```
knute-sifive, line 33: Did you intend for this code to be in the patch?

knute-sifive (commented):
Couple of things:
1. Can you add core mark to the regression (just like Dhrystone)?
2. Can you add a report that calculates the core mark performance?  (see https://github.com/riscv-software-src/riscv-perf-model/blob/master/reports/dhry_report.yaml)

traces/stf_trace_gen/coremark_riscv.patch
light2802: Yes, because coremark needs a few function definitions which are platform dependant.

light2802:
> Couple of things:
> 
>     1. Can you add core mark to the regression (just like Dhrystone)?
Which regression. Do you mean github workflow? 

>     2. Can you add a report that calculates the core mark performance?  (see https://github.com/riscv-software-src/riscv-perf-model/blob/master/reports/dhry_report.yaml)
Sure

traces/stf_trace_gen/coremark_riscv.patch
knute-sifive: Ok, where did the definitions originate?  Although the licensing is the same as Olympia's I do worry that if the author _changes_ that license and we are have this separate patch that does not align with that, there will be trouble.  Can you provide directions on how to get this patch instead of including it?

knute-sifive:
> 
> Which regression. Do you mean github workflow?
> 

Nope.  I mean as part of Olympia's regression target (which is automatically used by GH flows):

Here's an example:

https://github.com/riscv-software-src/riscv-perf-model/blob/e78fb2efe76d17f5e8172f1da0bf49358606d46c/test/sim/CMakeLists.txt#L37-L42

knute-sifive:
BTW, are you missing `reports/core_report.def`?

PR #67, opened 2023-07-11
Update stf patch

light2802:
Does not work yet. DO NOT MERGE


knute-sifive:
I know this is a draft, but seems to have the some of the same changes in PR #66

light2802:
> I know this is a draft, but seems to have the some of the same changes in PR #66
Yes because this patch is on top of  PR #66 . If felt to keep both separate as they are somewhat independant

light2802:
@knute-sifive  Can you help me with this? I cant figure out why it's not working. I tried to use gdb for debugging and basically, we never encounter an START_TRACE_OPC instruction. The same elf binary works with the old patch so its not an issue there. Thanks.

knute-sifive:
Finally had a chance to look at this.  This commit of dromajo broke us: https://github.com/chipsalliance/dromajo/commit/c197499682181fae46652bc9989e63b572d3c274

We can either apply the patch one revision before this one or figure out how this patch broke us.

knute-sifive (approved):

knute-sifive:
The latest Dromajo _hard-codes_ an instruction quantum to 10k (from just 1) which bypasses checks for trace start/stop markers.  This patch reduces the quantum back down to 1.  Eventually we will need to add trace start/stop detection in the main template of dromajo, but hopefully we'll have replaced before we need to do that.

Issue #69, opened 2023-07-19
Separate Address/Data Readiness


AaronGChan:
Separate address and data readiness in the LSU and scoreboard to allow for address calculations to run ahead if data is not ready. Currently we wait for both address and data operands to be ready before starting store operations.

AaronGChan:
Implemented in https://github.com/riscv-software-src/riscv-perf-model/pull/70.

@h0lyalg0rithm, the address/data separation is structured as follows:
In an `Inst` object, we have a data register mask and separate tracking in `RenameData`:
https://github.com/riscv-software-src/riscv-perf-model/blob/7c0b0513554a1622f05c9c3b27895f795e634960/core/Inst.hpp#L65
https://github.com/riscv-software-src/riscv-perf-model/blob/7c0b0513554a1622f05c9c3b27895f795e634960/core/Inst.hpp#L231

for the address tracking, tracking is done via the `src_reg_bit_mask`:
https://github.com/riscv-software-src/riscv-perf-model/blob/7c0b0513554a1622f05c9c3b27895f795e634960/core/Inst.hpp#L229

where it's stored in the integer rf, because address operands are always integer.

We always check the address operand for readiness first, and if it's ready, then the data for store operations. So ideally if address operand is ready, but data is not, you can runahead with the address calculation and the data will have a callback from the scoreboard when it's ready:
https://github.com/riscv-software-src/riscv-perf-model/blob/7c0b0513554a1622f05c9c3b27895f795e634960/core/LSU.cpp#L117

Let me know if you have any questions on this.

PR #70, opened 2023-07-20
Separation of Data/Address Tracking

AaronGChan:
Adds separate tracking address/data operands for stores (https://github.com/riscv-software-src/riscv-perf-model/issues/69).
Fixes isStore bug (https://github.com/riscv-software-src/riscv-perf-model/issues/78).
Fixes assertion bug in LSU.


knute-sifive:
Instead of using the `getSourceOpInfoList()` why not use the `RenameData` object that has the same information, but the register file already resolved.

core/LSU.cpp
```diff
@@ -111,13 +111,45 @@ namespace olympia
...
+            const auto & src_bits = inst_ptr->getSrcRegisterBitMask(reg_file);
+            if(scoreboard_views_[reg_file]->isSet(src_bits)){
+                operands_ready = true;
+            }
+            else{
+                scoreboard_views_[reg_file]->registerReadyCallback(src_bits, inst_ptr->getUniqueID(),
+                                    [this, inst_ptr](const sparta::Scoreboard::RegisterBitMask&)
+                                    {this->getInstsFromDispatch_(inst_ptr);});
+                ILOG("Registering Callback: " << inst_ptr);
+            }
```
knute-sifive, line 148: Perhaps you see a logic path that I missed, but how about this?  Yeah, it's a bit redundant (will check INT twice for example), but it's a good start...

```
bool all_ready = true; // assume all ready
for(const auto & src_reg : inst_ptr->getRenameData().getSourceList())
{
    const auto & src_bits = inst_ptr->getSrcRegisterBitMask(src_reg.rf);
    if(false == scoreboard_views_[src_reg.rf]->isSet(src_bits)) {
        all_ready = false;
        break;
    }
}

if (all_ready) {
    // Create load/store memory access info
...
```

AaronGChan: Yeah this looks a lot cleaner compared to what I have, will use your implementation instead and test.

knute-sifive:
> Should I create a separate scoreboard to track stores then?

You don't need a separate scoreboard, just a separate set of register bit masks in the instruction.  

For stores, there are two bit masks: one for address operands (reuse the existing `src_reg_big_masks_`, which are always register file INTEGER for loads/stores) and _new_ one for store data (register file variable).  

Start with an additional mask to Inst.hpp:
```
        RegisterBitMaskArray src_reg_bit_masks_;
        RegisterBitMaskArray dest_reg_bit_masks_;
        RegisterBitMask      store_data_mask_;
```
And in Rename, you ask the question: Are you a store?  If so, you set the bits in the `store_data_mask_`.

There's more to it than what I wrote, but should get you started...

core/LSU.cpp
AaronGChan: Implemented in [4189f59](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/4189f590b3157d70f90ed749a664447fd88080ea)

core/Inst.hpp
```diff
@@ -44,6 +44,9 @@ namespace olympia
             void setOriginalDestination(const Reg & destination){
                 original_dest_ = destination;
             }
+            void setAddrReg(const Reg & addr_reg){
+                addr_reg_ = addr_reg;
+            }
```
knute-sifive, line 49: Did you mean data register?  There is only 1 data register and multiple source registers for address.  I was thinking that the `src_` register list would be for address operands and `addr_reg_` would actually be `data_reg_`

core/LSU.cpp
```diff
@@ -111,13 +111,38 @@ namespace olympia
     // Receive new load/store instruction from Dispatch Unit
     void LSU::getInstsFromDispatch_(const InstPtr & inst_ptr)
     {
-        core_types::RegFile reg_file = core_types::RF_INTEGER;
-        const auto & srcs = inst_ptr->getSourceOpInfoList();
-        if(srcs.size() > 0){
-            reg_file = olympia::coreutils::determineRegisterFile(srcs[0]);
+        bool all_ready = true; // assume all ready
+        for(const auto & src_reg : inst_ptr->getRenameData().getSourceList())
```
knute-sifive, line 115: Yeah, this `for` loop would only check address operands, which are all INTEGER.  So all you need to do is
```
scoreboard_views_[core_types::RF_INTEGER]->isSet(inst_ptr->getSrcRegisterBitMask(core_types::RF_INTEGER);
```
and that's it.

Then, you check for data operands if (and only if) a store instruction.

core/Inst.hpp
AaronGChan: Not sure if I'm following of how we would have multiple source registers for address? For stores we would only ever have 1 data operand and 1 address operand?

From a coding standpoint I get why we should make src_ register the place for address operands because we can then always check src_registers for loads/stores like you explained in the other comment.

knute-sifive: Ugh, I keep mixing ARM/PowerPC ISA with RISC-V.  Indeed, there are no load/store instructions that use 2 integer sources for address calculation, just 1.

So, let me start over with my thinking reset.  🤣 

Yes, I think the `src_` register list should be only for address operands (sounds like you agree) and we should have a `data_src_reg_` for stores.  

> Also because we're using a separate mask for tracking address operands, do we then allocate a PRF for that address operand? 

Yes, you will always _refer_ to a PRF for the address sources just like any other register.  And yes, you will refer to a PRF for the store data just like any other register.  Those values usually come from an integer/float instruction like an `addi` or from a load (think memcpy).

In modern processors, MBRs are usually called something like store buffers or line fill buffers.  Either way, the memory to be stored comes from a register in the PRF.

AaronGChan: Addressed in [453b6c3](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/453b6c3b19892012263c7985f4af0fc4717b05d1)

core/LSU.cpp
AaronGChan: Addressed in [453b6c3](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/453b6c3b19892012263c7985f4af0fc4717b05d1)

core/Rename.cpp
```diff
@@ -132,10 +132,9 @@ namespace olympia
         }
 
         const auto & srcs = inst_ptr->getRenameData().getSourceList();
-        if(inst_ptr->isStoreInst() && srcs.size() < 2 && dests.size() == 0){
-            // free PRF for STORE instruction used for data tracking
-            core_types::RegFile rf = core_types::RF_INTEGER;
-            freelist_[rf].push(inst_ptr->getRenameData().getAddrReg().val);
+        if(inst_ptr->isStoreInst() && dests.size() == 0){
+            // decrement counter for store data tracking
+            --reference_counter_[inst_ptr->getRenameData().getDataReg().rf][inst_ptr->getRenameData().getDataReg().val];
```
knute-sifive, line 137: Wow, that's a big line!  😆 Can you break it up?
```suggestion
            const auto & data_reg = inst_ptr->getRenameData().getDataReg();
            --reference_counter_[data_reg.rf][data_reg.val];
```

core/Rename.cpp
```diff
@@ -177,12 +176,9 @@ namespace olympia
...
             // create an index count for each instruction entered
             const auto & dests = i->getDestOpInfoList();
-            if(dests.size() > 0 && i->isStoreInst() && i->getSourceOpInfoList().size() < 2){
-                ILOG("test1");
-            }
-            if(i->isStoreInst() && i->getSourceOpInfoList().size() < 2 && dests.size() == 0){
-                // For STOREs we count for INTEGER because addresses use INTEGER PRF
-                current_counts.cumulative_reg_counts[core_types::RF_INTEGER]++;
+            if(i->isStoreInst() && dests.size() == 0){
+                // For STOREs, addresses use INTEGER PRF
```
knute-sifive, line 180: So do loads.  😉

knute-sifive: Wait, this is the dest op list.  Stores do not have dests.  This should be an assert

core/Rename.cpp
```diff
@@ -272,33 +268,52 @@ namespace olympia
...
-                    auto & bitmask = renaming_inst->getSrcRegisterBitMask(rf);
-                    const uint32_t prf = map_table_[rf][num];
-                    reference_counter_[rf][prf]++;
-                    renaming_inst->getRenameData().setSource({prf, rf});
-                    bitmask.set(prf);
-
-                    ILOG("\tsetup source register bit mask "
-                         << sparta::printBitSet(bitmask)
-                         << " for '" << rf << "' scoreboard");
+                    if(renaming_inst->isStoreInst() && dests.size() == 0 && (src.field_id == mavis::InstMetaData::OperandFieldID::RS2)){
```
knute-sifive, line 273: Not sure why you're checking for dest list size here.  Stores will never have dests...

core/Rename.cpp
```diff
@@ -272,33 +268,52 @@ namespace olympia
...
 
-                    if(renaming_inst->isStoreInst() && srcs.size() < 2 && dests.size() == 0){
-                        // assigning a register from PRF for data register
-                        auto & addr_bitmask = renaming_inst->getAddrRegisterBitMask();
-                        core_types::RegFile rf = core_types::RF_INTEGER; // all address operations are integer
+                        ILOG("\tsetup store data register bit mask "
+                            << sparta::printBitSet(bitmask)
+                            << " for '" << rf << "' scoreboard");
+                    }
+                    else if(renaming_inst->isStoreInst() && dests.size() == 0){
```
knute-sifive, line 287: This statement can be combined with the previous `if` statement.  Here's the logic:

```
for (src : src_list)
{
    if( store instruction ) {
       if( src.pos == RS2 ) {
            setDataReg(src);
       }
        else {
            setSource(src);
        }
    }
}
```

core/Rename.cpp
AaronGChan: I check the dest list because isStoreInst() is true for both loads and stores:
https://github.com/riscv-software-src/riscv-perf-model/blob/f18ab54d6c7a1b5d5b058e58577ee3141ca2533c/core/Inst.hpp#L173

so I check `dests.size() == 0` to make sure it's a store

core/Rename.cpp
AaronGChan: Implemented in [e7fa84c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/e7fa84ca02af05a0422db09893bfd2815f6b2e11)

core/Rename.cpp
AaronGChan: Fix in [e7fa84c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/e7fa84ca02af05a0422db09893bfd2815f6b2e11)

I saw an increase in IPC (1.13 -> 1.4), just from fixing the logic behind isStoreInst, not sure if there was another bug induced by that fix, so might need additional checking.

core/Rename.cpp
AaronGChan: Fixed in [e7fa84c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/e7fa84ca02af05a0422db09893bfd2815f6b2e11)

core/Inst.hpp
```diff
@@ -205,13 +219,15 @@ namespace olympia
         uint64_t               unique_id_     = 0; // Supplied by Fetch
         uint64_t               program_id_    = 0; // Supplied by a trace Reader or execution backend
         bool                   is_speculative_ = false; // Is this instruction soon to be flushed?
+        bool                   is_store_ = opcode_info_->getDestOpInfoList().size() == 0 && inst_arch_info_->isLoadStore();
```
knute-sifive, line 222: Actually Mavis has the information you're looking for directly:

https://github.com/sparcians/mavis/blob/c3c1060cf59baf8c16c13ddd2473f84e7c3aeb65/mavis/OpcodeInfo.h#L284C14-L287

Change this variable to a `const bool is_store_;` and do this in the _constructor_ of the instruction:
```
        Inst(const mavis::OpcodeInfo::PtrType& opcode_info,
             const InstArchInfo::PtrType     & inst_arch_info,
             const sparta::Clock             * clk) :
            opcode_info_    (opcode_info),
            inst_arch_info_ (inst_arch_info),
            is_store_(opcode_info->isInstType(mavis::OpcodeInfo::InstructionTypes::STORE),
            status_state_(Status::FETCHED)
...
```

core/LSU.cpp
```diff
@@ -10,43 +10,35 @@ namespace olympia
...
+                                                                 ldst_inst_queue_("lsu_inst_queue", p->ldst_inst_queue_size, getClock()),
+                                                                 ldst_inst_queue_size_(p->ldst_inst_queue_size),
+
+                                                                 tlb_always_hit_(p->tlb_always_hit),
+                                                                 dl1_always_hit_(p->dl1_always_hit),
+
+                                                                 issue_latency_(p->issue_latency),
+                                                                 mmu_latency_(p->mmu_latency),
+                                                                 cache_latency_(p->cache_latency),
+                                                                 complete_latency_(p->complete_latency)
```
knute-sifive, line 25: Um... hate the spacing.  😆 

To be fair, we need to put together some coding standards for this project.

core/LSU.cpp
```diff
@@ -109,22 +98,48 @@ namespace olympia
...
     // Receive new load/store instruction from Dispatch Unit
-    void LSU::getInstsFromDispatch_(const InstPtr & inst_ptr)
+    void LSU::getInstsFromDispatch_(const InstPtr &inst_ptr)
     {
-        core_types::RegFile reg_file = core_types::RF_INTEGER;
-        const auto & srcs = inst_ptr->getSourceOpInfoList();
-        if(srcs.size() > 0){
-            reg_file = olympia::coreutils::determineRegisterFile(srcs[0]);
+        bool all_ready = true; // assume all ready
+        if (!scoreboard_views_[core_types::RF_INTEGER]->isSet(inst_ptr->getSrcRegisterBitMask(core_types::RF_INTEGER)))
```
knute-sifive, line 104: Pls move your comment that you are checking address operands above the `if` statement

core/LSU.cpp
```diff
@@ -109,22 +98,48 @@ namespace olympia
...
+            const auto &src_bits = inst_ptr->getSrcRegisterBitMask(core_types::RF_INTEGER);
+            scoreboard_views_[core_types::RF_INTEGER]->registerReadyCallback(src_bits, inst_ptr->getUniqueID(),
+                                                                             [this, inst_ptr](const sparta::Scoreboard::RegisterBitMask &)
+                                                                             {
+                                                                                 this->getInstsFromDispatch_(inst_ptr);
+                                                                             });
+            ILOG("Instruction NOT ready: " << inst_ptr << " Bits needed:" << sparta::printBitSet(src_bits));
+        }
+
+        if (inst_ptr->isStoreInst())
```
knute-sifive, line 117: Pls add a comment above the `if` statement that you're checking for data ready for stores.

core/Rename.cpp
```diff
@@ -266,30 +267,62 @@ namespace olympia
...
-                    const uint32_t prf = map_table_[rf][num];
-                    reference_counter_[rf][prf]++;
-                    renaming_inst->getRenameData().setSource({prf, rf});
-                    bitmask.set(prf);
-                    ILOG("\tsetup source register bit mask "
-                         << sparta::printBitSet(bitmask)
-                         << " for '" << rf << "' scoreboard");
-                }
+                    if(renaming_inst->isLoadStoreInst()){
+                        if(src.field_id == mavis::InstMetaData::OperandFieldID::RS2){
```
knute-sifive, line 275: Pls add comments on what this code is doing.

core/Rename.cpp
```diff
@@ -266,30 +267,62 @@ namespace olympia
...
+                                << sparta::printBitSet(bitmask)
+                                << " for '" << rf << "' scoreboard");
+                        }
+                        else{
+                            core_types::RegFile rf = core_types::RegFile::RF_INTEGER;
+                            auto & bitmask = renaming_inst->getSrcRegisterBitMask(rf);
+                            const uint32_t prf = freelist_[rf].front();
+                            freelist_[rf].pop();
+                            renaming_inst->getRenameData().setSource({prf, rf});
+                            reference_counter_[rf][prf]++;
```
knute-sifive, line 294: Wait, isn't this a source register?  Why are you grabbing a register from the FL?

knute-sifive:
Did you run a linter over this code?  A lot of style changes, which kinda pollutes the review.  Also, Suraj and team are working on this code and these changes might cause massive conflicts with their work.  Suggest that you do style changes in a separate PR in the future...

core/Rename.cpp
AaronGChan: I was thinking we needed to hold a scoreboard space to allow for address calculation checking to be tracked, so I assumed we needed to allocate a PRF to "hold" that bit. But I realized I went too far ahead in my thinking and created a bug in the process, because I'm always setting that bit to ready and not actually checking if the address is ready. I'll fix/change it back to only source renaming checking/mapping.

Just to make sure we're on the same page, the goal of the address/data separation is only meant to create separate data structures for tracking either address or data, and then Suraj and the team will handle adding in a way to track the address calculation if it runs ahead?

core/LSU.cpp
AaronGChan: Ah sorry, will fix. I had copied my code over from another branch, but didn't realize the spacing was different.

AaronGChan:
> Did you run a linter over this code? A lot of style changes, which kinda pollutes the review. Also, Suraj and team are working on this code and these changes might cause massive conflicts with their work. Suggest that you do style changes in a separate PR in the future...

I haven't used a linter on this code, will make sure to use one going forward.

knute-sifive:
> I haven't used a linter on this code, will make sure to use one going forward.

Sorry, not suggesting that you use one -- I was asking _if_ you used one.  There were a LOT of syntax changes that looked like you ran one.

core/Rename.cpp
knute-sifive: That's correct.  The address is *only* calculated in the LS and usually store alongside the operation in either a load queue or a store queue.  It will *never* write that value to a PRF -- no reason for that.

AaronGChan: Fixed in [5212a4c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/5212a4c3274e758abca7bebf8ebc45a4ac210ca9)

core/Rename.cpp
AaronGChan: Comments added in [5212a4c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/5212a4c3274e758abca7bebf8ebc45a4ac210ca9)

core/LSU.cpp
AaronGChan: Comments added in [5212a4c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/5212a4c3274e758abca7bebf8ebc45a4ac210ca9)

core/LSU.cpp
AaronGChan: Comments added in [5212a4c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/5212a4c3274e758abca7bebf8ebc45a4ac210ca9)

core/LSU.cpp
AaronGChan: Spacing fixed in [5212a4c](https://github.com/riscv-software-src/riscv-perf-model/pull/70/commits/5212a4c3274e758abca7bebf8ebc45a4ac210ca9)

AaronGChan:
@knute-sifive,
For the following RAW hazard,
```
{
        "mnemonic": "fadd.d",
        "fs1": 10,
        "fs2": 17,
        "fd": 3
    },
    {
        "mnemonic": "fsw",
        "fs2": 2,
        "fs1": 3,
        "imm": 4
}
```
We've discussed that the address operands are always integer, but if the address is coming from a floating point register as in the above hazard, we have to track the floating point scoreboard in order to properly observe and stall for the floating point destination of instruction 1.
In the current code, we assume the address operand is stored in integer register file, so we check only this:
https://github.com/riscv-software-src/riscv-perf-model/blob/5212a4c3274e758abca7bebf8ebc45a4ac210ca9/core/LSU.cpp#L117
which would not observe the RAW properly and not stall. So shouldn't we not mark the address operand of a store/load as always integer, but instead track it based on the operand type? Not sure if I misinterpreted our past discussions on this and implemented it wrong.

PR #77, opened 2023-08-14
Seperate the MMU and cache from the lsu

h0lyalg0rithm:
I kept the logic the same as it was earlier.
One of the design considerations was to use the BIU to communicate with the cache and MMU.However it was later decided to keep the communication simple until we design a more robust MMU design


knute-sifive (approved):
LGTM

core/LSU.cpp
```diff
@@ -257,7 +263,7 @@ namespace olympia
         }
 
         // Access TLB, and check TLB hit or miss
-        bool TLB_HIT = MMULookup_(mem_access_info_ptr);
+        bool TLB_HIT = mmu_->Lookup(mem_access_info_ptr);
```
knute-sifive, line 266: Two things:
1. While we haven't really established coding style guidelines, this variable name should not be in all caps.  Request `tlb_hit` to match the other code
2. This variable can be `const` : `const bool tlb_hit =`.  This is helpful is avoiding accidental assignments when you intended to compare.

core/LSU.cpp
```diff
@@ -353,22 +359,22 @@ namespace olympia
 
         if (cacheBypass) {
```
knute-sifive, line 360: Can you change the variable name to be consistent? `cache_bypass`

core/LSU.cpp
```diff
@@ -353,22 +359,22 @@ namespace olympia
...
             if (isAlreadyHIT) {
```
knute-sifive, line 361: Same here.

core/LSU.cpp
```diff
@@ -353,22 +359,22 @@ namespace olympia
...
             else {
-                sparta_assert(false, "Cache access is bypassed without a valid reason!");
+                sparta_assert(false, "DCache access is bypassed without a valid reason!");
             }
             return;
         }
 
         // Access cache, and check cache hit or miss
-        const bool CACHE_HIT = cacheLookup_(mem_access_info_ptr);
+        const bool CACHE_HIT = data_cache_->cacheLookup(mem_access_info_ptr);
```
knute-sifive, line 377: Variable name change, please.

core/LSU.cpp
```diff
@@ -596,6 +608,36 @@ namespace olympia
...
+    void LSU::popLoadQueue_(const InstPtr & inst_ptr)
+    {
+        // Look for the instruction to be completed, and remove it from load queue
+        for (auto iter = load_queue_.begin(); iter != load_queue_.end(); iter++) {
+            if ((*iter)->getInstPtr() == inst_ptr) {
+                load_queue_.erase(iter);
+
+                return;
+            }
+        }
```
knute-sifive, line 621: A faster way to do this: when the load is allocated into the load queue, store the index (iterator) of its position in the request.  Then, when popping, check that iterator's validity and if valid, just erase it.

core/LSU.cpp
```diff
@@ -596,6 +608,36 @@ namespace olympia
...
+    void LSU::popStoreQueue_(const InstPtr & inst_ptr)
+    {
+        // Look for the instruction to be completed, and remove it from store queue
+        for (auto iter = store_queue_.begin(); iter != store_queue_.end(); iter++) {
+            if ((*iter)->getInstPtr() == inst_ptr) {
+                store_queue_.erase(iter);
+
+                return;
+            }
+        }
```
knute-sifive, line 636: Same suggestion here as above.

core/LSU.cpp
```diff
@@ -643,43 +685,28 @@ namespace olympia
...
+    void LSU::appendLoadQueue_(const LoadStoreInstInfoPtr & inst_info_ptr)
     {
-        const InstPtr & inst_ptr = mem_access_info_ptr->getInstPtr();
-        uint64_t vaddr = inst_ptr->getTargetVAddr();
+        sparta_assert(load_queue_.size() <= ld_queue_size_,
+                      "Appending load queue causes overflows!");
 
-        bool tlb_hit = false;
+        // Always append newly dispatched instructions to the back of load queue
+        load_queue_.push_back(inst_info_ptr);
```
knute-sifive, line 695: Yeah, you can do this:
```suggestion
        inst_info_ptr->storeLQPosition(load_queue_.push_back(inst_info_ptr));
```

core/LSU_MMU_Shared.hpp
knute-sifive, line 1: Why is called "shared"?

core/LSU_MMU_Shared.hpp
```diff
@@ -0,0 +1,125 @@
...
+            return mmu_access_state_;
+        }
+
+        void setMMUState(const MMUState &state) {
+            mmu_access_state_ = state;
+        }
+
+        const CacheState &getCacheState() const {
+            return cache_access_state_;
+        }
```
knute-sifive, line 70: MMUState and CacheState are enums (POD types).  Returning a reference converts them to a pointer and de dereferenced.  Makes sense to just return by value.   (It's a good chance the compiler will do that for you)

core/MMU.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    bool MMU::Lookup(const MemoryAccessInfoPtr &mem_access_info_ptr) {
+        const InstPtr &inst_ptr = mem_access_info_ptr->getInstPtr();
+        uint64_t vaddr = inst_ptr->getTargetVAddr();
+
+        bool tlb_hit = false;
+
+        // C++ comma operator: assign tlb_hit first, then evaluate it. Just For Fun
+        if (tlb_hit = tlb_always_hit_, tlb_hit) {
+        } else {
```
knute-sifive, line 24: Might as well convert to 
```suggestion
        bool tlb_hit = tlb_always_hit_;
        if(false == tlb_hit) {
             // miss code
```

core/MMU.cpp
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+
+
+        if (tlb_always_hit_) {
+            ILOG("TLB HIT all the time: vaddr=0x" << std::hex << vaddr);
+            tlb_hits_++;
+        } else if (tlb_hit) {
+            ILOG("TLB HIT: vaddr=0x" << std::hex << vaddr);
+            tlb_hits_++;
```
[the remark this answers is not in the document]
h0lyalg0rithm: I think this was more for debugging purposes

core/LSU_MMU_Shared.hpp
h0lyalg0rithm: I though I might have to put in more structures apart from the memoryaccessinfo.
But it makes sense to rename it to `MemoryAccessInfo.hpp`

core/LSU.cpp
h0lyalg0rithm: I will remove the code from this PR to keep the changes related to the separation of the cache and mmu

h0lyalg0rithm:
@knute-sifive Both Vineeth and I have completed this PR.We would use the final commit to branch out and work on our changes to the lsu and cache

core/LSU.cpp
```diff
@@ -925,6 +612,64 @@ namespace olympia
...
+        // Check if flushing event occurred just now
+        if (!ldst_pipeline_.isValid(stage_id)) {
+            return;
+        }
+
+        const MemoryAccessInfoPtr & mem_access_info_ptr = ldst_pipeline_[stage_id];
+        ILOG(mem_access_info_ptr);
+
+        bool isAlreadyHIT = (mem_access_info_ptr->getMMUState() == MemoryAccessInfo::MMUState::HIT);
+        bool MMUBypass = isAlreadyHIT;
```
knute-sifive, line 632: Duplicate `bool` variables.  Remove one; make the other `const` and change the variable name to match the rest of the code

core/LSU.cpp
```diff
@@ -925,6 +612,64 @@ namespace olympia
...
+
+    // Handle MMU access request
+    void LSU::handleMMULookupReq2_()
+    {
+        if(!mmu_hit_)
+            ldst_pipeline_.invalidateStage(static_cast<uint32_t>(PipelineStage::MMU_LOOKUP));
+    }
+
+    void LSU::getInstFromMMU_(const MemoryAccessInfoPtr &memory_access_info_ptr) {
+        auto inst_ptr = memory_access_info_ptr->getInstPtr();
```
knute-sifive, line 652: `const auto &` to avoid the copy

Issue #78, opened 2023-08-17
All load_store instructions are marked as store Instructions


h0lyalg0rithm:
All the instructions which are load or store always return true when this function is called https://github.com/riscv-software-src/riscv-perf-model/blob/620b664797585beb6b42d3a1b004b708072d0f65/core/Inst.hpp#L173
The value for `isLoadStore()` is true if the instruction is either load or store.
https://github.com/riscv-software-src/riscv-perf-model/blob/620b664797585beb6b42d3a1b004b708072d0f65/core/InstArchInfo.hpp#L91

AaronGChan:
Fixed in https://github.com/riscv-software-src/riscv-perf-model/pull/70.

PR #79, opened 2023-08-17
Fix the isStoreInst function

h0lyalg0rithm:
Fixes issue #78 
This change would affect the behavior of the LSU


h0lyalg0rithm:
Fixed in #70

PR #82, opened 2023-08-26
Setup conda cache and ccache

h0lyalg0rithm:
The goal of the PR is to improve build times.
Currently the build steps take around  > ~15 mins and most of the time is spent building sparta and the other dependencies.
This should bring it down to roughly ~6 mins.

CCache is used to cache build files, the cache for each branch is unique to prevent cache invalidation issues between different branches.

Conda dependencies are also cached although the time to build is not affected as much roughly  ~1 mins but would be higher if the network is slow.Conda spends a lot of time checking the dependencies.

### DIFF
--- a/.github/workflows/ubuntu-build.yml
+++ b/.github/workflows/ubuntu-build.yml
@@ -44,9 +44,26 @@ jobs:
         with:
           python-version: 3.8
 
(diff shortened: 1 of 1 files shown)


knute-sifive (approved):
Very nice!

PR #85, opened 2023-09-01
In order load-store pipeline

h0lyalg0rithm:
This PR includes the changes from https://github.com/riscv-software-src/riscv-perf-model/pull/77 .
I have implemented stall_on_pipeline_miss as a parameter for the LSU.
It assumes the timing from MMU and Cache to be as follows

Lookup => 1 cycle
Ready => N cycles

The load store unit would execute the instructions one after the other and stall the pipeline if it encounters a miss.


h0lyalg0rithm:
I have included a new file test/core/dispatch/test_cores/test_small_core_stall.yaml which runs the existing tests with the stall_on_pipeline_miss parameter active

core/LSU.cpp
```diff
@@ -57,16 +59,16 @@ namespace olympia
...
-        ldst_pipeline_.registerHandlerAtStage<sparta::SchedulingPhase::PostTick>(static_cast<uint32_t>(PipelineStage::MMU_LOOKUP),
+        ldst_pipeline_.registerHandlerAtStage<sparta::SchedulingPhase::Tick>(static_cast<uint32_t>(PipelineStage::MMU_LOOKUP),
                                               CREATE_SPARTA_HANDLER(LSU, handleMMULookupReq2_));
 
-        ldst_pipeline_.registerHandlerAtStage<sparta::SchedulingPhase::Update>(static_cast<uint32_t>(PipelineStage::CACHE_LOOKUP),
+        ldst_pipeline_.registerHandlerAtStage<sparta::SchedulingPhase::Tick>(static_cast<uint32_t>(PipelineStage::CACHE_LOOKUP),
                                               CREATE_SPARTA_HANDLER(LSU, handleCacheLookupReq1_));
 
-        ldst_pipeline_.registerHandlerAtStage<sparta::SchedulingPhase::PostTick>(static_cast<uint32_t>(PipelineStage::CACHE_LOOKUP),
+        ldst_pipeline_.registerHandlerAtStage<sparta::SchedulingPhase::Tick>(static_cast<uint32_t>(PipelineStage::CACHE_LOOKUP),
```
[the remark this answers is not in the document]
h0lyalg0rithm: I updated the PR  with the changes from the other branch.
It should remove the multiple handles on the same pipe stage

core/LSU.cpp
```diff
@@ -273,9 +280,10 @@ namespace olympia
         out_cache_lookup_req_.send(mem_access_info_ptr);
     }
 
-    void LSU::getInstFromCache_(const MemoryAccessInfoPtr &memory_access_info_ptr)
-    {
-        auto inst_ptr = memory_access_info_ptr->getInstPtr();
+    // Ready signal from the MMU
+    void LSU::getInstFromCache_(const MemoryAccessInfoPtr &memory_access_info_ptr) {
```
knute-sifive, line 284: This function name is confusing to me.  For a split second I thought I was looking at Fetch!  😆 

The function is not getting any insts from the cache -- it's getting a MMU acknowledgment.  Might as well name it as such..

core/LSU.cpp
```diff
@@ -482,9 +505,15 @@ namespace olympia
                 continue;
             }
 
-            // Pick winner
-            if (!(*win_ptr_iter)->isReady() || (*iter)->winArb(*win_ptr_iter)) {
-                win_ptr_iter = iter;
+            if (stall_pipeline_on_miss_) {
+                if((*iter)->getPriority() == LoadStoreInstInfo::IssuePriority::NEW_DISP) {
+                    return *iter;
+                }
```
knute-sifive, line 511: Perhaps you can point me to the docs for this, but if we are stalling the pipeline on a miss, shouldn't we check to see if the pipeline _is stalled_ before picking new op?  Or are you **only** looking for new ops _because_ of the stalling scenario?

h0lyalg0rithm: @knute-sifive This PR only implements stall on miss, a new instruction is issued only when an instruction exits the pipeline.This is why I am look for instruction which are ready and newly dispatched.

core/LSU.cpp
h0lyalg0rithm: Yeah makes sense. Fixing it now

h0lyalg0rithm:
@knute-sifive @arupc I think I implemented something totally different here.
I implemented an in order load-store execute instead of a stall on pipeline miss behaviour.
I will change the title and the description to match that

PR #90, opened 2023-09-27
Ensure Queue is empty before finishing simulation

h0lyalg0rithm:


h0lyalg0rithm:
@knute-sifive I added this assert to make sure there are no pending stores and the current model fails.Do you some ideas on this

knute-sifive:
I ran the regression with your branch and the failure makes sense -- and doesn't.  😄 

One of the tests terminates simulation after 1M instructions have retired:
```
olympia -i 1M --workload traces/core_riscv.zstf
```
which will leave "unretired" stores in the ROB.  In other words -- that's ok!  

This test works fine:
```
olympia -i 1M --workload traces/core_riscv.zstf
```

h0lyalg0rithm:
Got it, should I change the test cases to remove the instruction limit.

knute-sifive:
Well, you can remove the instruction limit if you want, but the simulator _should not die_ if a user wants to only generate a pipeout or report or whatever for just the first few million instructions.  

In other words, I do not think this assert is required.

h0lyalg0rithm:
Got it, Is there another way to get around this issue?
Its pretty useful when developing to know that all the instructions were correctly handled in the LSU

PR #91, opened 2023-10-04
Non Blocking Cache Implementation

vineethkm:
Modelling the cache behaviour as mentioned in the discussion #71


core/DCache.cpp
```diff
@@ -48,9 +66,9 @@ namespace olympia {
             cache_hit = (cache_line != nullptr) && cache_line->isValid();
 
             // Update MRU replacement state if DCache HIT
-            if (cache_hit) {
-                l1_cache_->touchMRU(*cache_line);
-            }
+            //if (cache_hit) {
+            //    l1_cache_->touchMRU(*cache_line);
+            //}
```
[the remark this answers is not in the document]
vineethkm: I initially thought of creating a function for updating replacement info. I have removed the comment.

core/DCache.cpp
```diff
@@ -87,37 +87,65 @@ namespace olympia {
...
-            cache_pipeline_.append(memory_access_info_ptr);
+            cache_pipeline_.append(memory_access_info_ptr); // Data read stage executes in the next cycle
         }
     }
 
     // Incoming Cache Refill request from BIU
-    void DCache::getAckFromBIU_(const InstPtr &inst_ptr) {
+    // Precedence: Before S1 (Cache Lookup)
+    void DCache::getRefillFromBIU_(const InstPtr &inst_ptr) {
+        // change inst_ptr to cache_line TODO
```
knute-sifive, line 122: `inst_ptr` is unused.  Consider changing `incoming_cache_refill_` from a bool to the instruction (set it to `nullptr` if none) being serviced for the refill and/or use the instruction pointer in an assert to make sure the `current_refill_mshr_entry_` matches the instruction.

core/DCache.cpp
```diff
@@ -87,37 +87,65 @@ namespace olympia {
...
-        // TODO
+        
+        // Write cache line to line fill buffer of ongoing mshr entry refill request
+        (*current_refill_mshr_entry_)->setDataArrived(true);
+
+        // Proceed to next stage (next cycle)
+        uev_cache_refill_write_.schedule(1);
+    }
+
+    void DCache::incomingRefillBIU_() {
```
knute-sifive, line 132: How is the method different from the one above?

core/DCache.cpp
```diff
@@ -212,8 +246,40 @@ namespace olympia {
...
+            return;
+        }
+
+        // Requests are sent in FIFO order
+        current_refill_mshr_entry_ = mshr_file_.begin();
+
+        // Send BIU request
+        // out_biu_req_.send(cache_line);
+        // Temp fix: call event after fixed time (4 cycles)
+        uev_biu_incoming_refill_.schedule(4);
```
knute-sifive: What's the fix for?  What are your thoughts on the real design here?

core/DCache.hpp
```diff
@@ -91,6 +105,7 @@ namespace olympia {
         const bool l1_always_hit_;
         bool busy_;
         uint32_t cache_latency_;
+        uint64_t cache_line_size_;
         uint32_t max_mshr_entries_;
```
knute-sifive: These variables are uninitialized.

core/DCache.cpp
vineethkm: I have decoupled the DCache from the BIU. Instead of using the port connected to the BIU, I would schedule an event to simulate the cache refill request.

core/DCache.cpp
vineethkm: > I have decoupled the DCache from the BIU. Instead of using the port connected to the BIU, I would schedule an event to simulate the cache refill request.

core/DCache.cpp
knute-sifive: Oh, ok.  So from today's SIG meeting, in _another_ PR, you'll be updating the BIU/Cache interface and reattach the BIU.

core/DCache.hpp
vineethkm: I initialize them in the DCache constructor

core/DCache.cpp
vineethkm: Yes

core/DCache.cpp
vineethkm: Understood, I will make the changes

core/DCache.hpp
knute-sifive: Ah, you're missing my point.  😄   You should _ALWAYS_ provide a default value when declaring a variable _just in case_ it's not or forgotten in the constructor.

If you do not want to provide a default value, then make them `const` which will _force_ you to provide a value in the constructor.  In fact -- please make the `const` now so you have them initialized to constant values.  
Thanks!

vineethkm: Understood, I have made those changes now

klingaard (commented):
This is looking good -- what's the plan for integration?

vineethkm:
> This is looking good -- what's the plan for integration?

The tests are running successfully with the LSU. I'm working on adding a more comprehensive test. The BIU Connection and Data Write Buffer will be added in a subsequent PR.

PR #92, opened 2023-10-04
Lsu allow spec load exec

h0lyalg0rithm:
This PR introduces the following
- Allows loads to perform Virtual address to Physical address translation before older loads complete.
  This can be configured using a parameter in the LSU
- Allow non blocking cache lookup requests. Depends on #91 for the cache to support this feature
- Implements Ready queue - Simulator only structure to speed up instruction lookup in the LSU.
- Mitigates data hazards introduced by instructions running out of order
- The length of the different stages of the LSU pipeline can be configured through the parameters.


arupc (commented):
Can we add some documentation on what are the purpose of various in and out ports of LSU, at least between LSU & MMU and LSU & Dache

LSU / MMU interface has 3 input ports and 1 output:
* output:
   - out_mmu_lookup_req
 * inputs:
   - in_mmu_lookup_req
   - in_mmu_lookup_ack
   - in_mmu_free_req
  
Similary, LSU / Dache interface has 3 input ports and 1 output port:
* output:
   - out_cache_lookup_req
 * inputs:
   - in_cache_lookup_req
   - in_cache_lookup_ack
   - in_cache_free_req

core/LSU.cpp
```diff
@@ -452,12 +679,218 @@ namespace olympia
...
+        }
 
+        sparta_assert(false, "Cannot replay a non existing instruction " << inst_ptr);
+    }
 
     ////////////////////////////////////////////////////////////////////////////////
     // Regular Function/Subroutine Call
     ////////////////////////////////////////////////////////////////////////////////
 
+    bool LSU::instPresentInQueues(const InstPtr &inst_ptr)
```
arupc, line 731: Please add documentation (even if just one line) stating what each method is doing.

core/LSU.hpp
```diff
@@ -319,19 +339,55 @@ namespace olympia
...
+
         ////////////////////////////////////////////////////////////////////////////////
         // Regular Function/Subroutine Call
         ////////////////////////////////////////////////////////////////////////////////
 
+        // Check if instruction is present in the any of the queues
+        bool instPresentInQueues(const InstPtr &inst_ptr);
+
+        // Check if the instruction is present in a specific queue
+        bool instInQueue(LoadStoreIssueQueue &queue,const InstPtr &inst_ptr);
```
arupc, line 364: what do we need both `instPresentInQueues` and `instInQueue`? 

Also, by "queue", do we mean just the issue queue (as opposed to load queue, store queue)? If so, may be change the comment to reflect that.

core/LSU.hpp
```diff
@@ -319,19 +339,55 @@ namespace olympia
...
+
+        bool allOlderStoresIssued(const InstPtr &inst_ptr);
+
+        bool olderStoresIssued(LoadStoreIssueQueue &queue, const InstPtr &inst_ptr);
+
+        void readyDependentLoads(const LoadStoreInstInfoPtr &);
+
+        void abortYoungerLoads(const MemoryAccessInfoPtr &);
+
+        void deallocateYoungerLoadFromQueue(LoadStoreIssueQueue &, const InstPtr &);
```
arupc, line 377: Please add one line comment describing the purpose of each the above methods

core/LSU.cpp
```diff
@@ -122,6 +153,12 @@ namespace olympia
     // Receive new load/store instruction from Dispatch Unit
     void LSU::getInstsFromDispatch_(const InstPtr &inst_ptr)
     {
+        if(!instPresentInQueues(inst_ptr)){
```
arupc, line 156: Is this condition ever going to be false? If this condition is false, then does it mean, dispatch sent an instruction which is already there in the ldst issue queue?

core/LSU.hpp
```diff
@@ -319,19 +339,55 @@ namespace olympia
...
+
+        // Remove instrunction from pipeline which share the same address
+        void invalidatePipeline(const InstPtr &);
```
arupc, line 380: All private methods should be  suffixed with underscore, following the convention of the rest of the code and to enhance the readability of the code

arupc (commented):
Can you please explain what these priorities / states mean and when they would occur? Not sure I understand what is meant by 'mss ack' and 'another oustanding miss finish' 

```
                HIGHEST = 0,
                __FIRST = HIGHEST,
                CACHE_RELOAD,   // Receive mss ack, waiting for cache re-access
                CACHE_PENDING,  // Wait for another outstanding miss finish
                MMU_RELOAD,     // Receive for mss ack, waiting for mmu re-access
                MMU_PENDING,    // Wait for another outstanding miss finish
                NEW_DISP,       // Wait for new issue
                LOWEST,
```

Does `NEW_DISP` simply mean dispatched and waiting to be issued?

arupc (commented):
One drawback of having a enumeration of IssuePriority is that it _somewhat_ ties us to having one  issue priority policy. If we want to allow multiple issue priority policies as a paramter, it may be difficult to with IssuePriority enumerator.

For that, we should keep the CACHE/MMU_RELOAD/PENDING info as states of the LoadStoreInfo class and allow different policies to decide which one to issue.

This can be done in a different PR after this is merged in.

core/LSU.cpp
h0lyalg0rithm: Yes I will add them

knute-sifive (dismissed):
This is a great start.  A few items here and there to tackle, obviously

core/LSU.hpp
```diff
@@ -277,9 +285,17 @@ namespace olympia
...
-        LoadStorePipeline ldst_pipeline_
-            {"LoadStorePipeline", static_cast<uint32_t>(PipelineStage::NUM_STAGES), getClock()};
+        using LoadStorePipeline = sparta::Pipeline<LoadStoreInstInfoPtr>;
+        LoadStorePipeline ldst_pipeline_;
+
+        uint32_t ADDRESS_CALCULATION = 0;
+        uint32_t MMU_LOOKUP = 1;
+        uint32_t CACHE_LOOKUP = 2;
+        uint32_t CACHE_READ = 3;
+        uint32_t COMPLETE = 4;
```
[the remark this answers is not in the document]
h0lyalg0rithm: @knute-sifive I use these variables as an alias for the pipeline stage. In the constructor I would replace them based on the length of the pipeline the user sets as a parameter.

core/LSU.hpp
```diff
@@ -49,6 +49,14 @@ namespace olympia
...
             // Parameters for ldst_inst_queue
             PARAMETER(uint32_t, ldst_inst_queue_size, 8, "LSU ldst inst queue size")
+            PARAMETER(uint32_t, replay_buffer_size, ldst_inst_queue_size, "Replay buffer size")
+            PARAMETER(uint32_t, replay_issue_delay, 3, "Replay Issue delay")
+            // LSU microarchitecture parameters
+            PARAMETER(bool, allow_speculative_load_exec, true, "Allow loads to proceed speculatively before all older store addresses are known")
+            // Pipeline length
+            PARAMETER(uint32_t, mmu_lookup_stage_length, 1, "Length of the mmu lookup stage")
+            PARAMETER(uint32_t, cache_lookup_stage_length, 1, "Length of the cache lookup stage")
+            PARAMETER(uint32_t, cache_read_stage_length, 1, "Length of the cache read stage")
```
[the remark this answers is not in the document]
h0lyalg0rithm: I designed the parameters to be in setup in a ldst_pipeline agnostic way
By default these are the mappings

| Stage         | Pipeline Alias     | Default |
|--------------|-----------|----------- |
| ADDRESS_GENERATION | 0      | 0 |
| MMU_LOOKUP      |  mmu_lookup_stage | 1 |
| CACHE_LOOKUP      |  mmu_lookup_stage + cache_lookup_stage | 2 |
| CACHE_READ      |  mmu_lookup_stage + cache_lookup_stage + cache_read_stage | 3 |
| COMPLETE      |  mmu_lookup_stage + cache_lookup_stage + cache_read_stage + 1 | 4 |

This would ensure that the stages are in the order without checking for the overlaps as the arithmetic would handle it.I should add validation to ensure that the length parameters are atleast 1.

example:
```
mmu_lookup_stage_length => 1
cache_lookup_stage_length => 2
cache_read_stage_length =>  2
```
| Stage         | Pipeline Alias     | Value |
|--------------|-----------| -----------|
| ADDRESS_GENERATION | 0      | 0  |
| MMU_LOOKUP      |  mmu_lookup_stage | 1|
| CACHE_LOOKUP      |  mmu_lookup_stage + cache_lookup_stage | 3|
| CACHE_READ      |  mmu_lookup_stage + cache_lookup_stage + cache_read_stage | 5 |
| COMPLETE      |  mmu_lookup_stage + cache_lookup_stage + cache_read_stage + 1 | 6 |

core/LSU.cpp
h0lyalg0rithm: Yes this condition can be false. Currently there is a callback that is registered in the same method.
One for one of operands and other for the data 
https://github.com/riscv-software-src/riscv-perf-model/pull/92/files/11626bef0bcdc815dd6cc9b37dbe7f2d6d558858#diff-f42bfd79bf07a70c3e64bbe0c640480fd5ee792c0c0465f138d9bad9219959a0R183

https://github.com/riscv-software-src/riscv-perf-model/pull/92/files/11626bef0bcdc815dd6cc9b37dbe7f2d6d558858#diff-f42bfd79bf07a70c3e64bbe0c640480fd5ee792c0c0465f138d9bad9219959a0R170

h0lyalg0rithm:
> Can you please explain what these priorities / states mean and when they would occur? Not sure I understand what is meant by 'mss ack' and 'another oustanding miss finish'
> 
> ```
>                 HIGHEST = 0,
>                 __FIRST = HIGHEST,
>                 CACHE_RELOAD,   // Receive mss ack, waiting for cache re-access
>                 CACHE_PENDING,  // Wait for another outstanding miss finish
>                 MMU_RELOAD,     // Receive for mss ack, waiting for mmu re-access
>                 MMU_PENDING,    // Wait for another outstanding miss finish
>                 NEW_DISP,       // Wait for new issue
>                 LOWEST,
> ```
> 
> Does `NEW_DISP` simply mean dispatched and waiting to be issued?
The priority values are used to select the instruction that needs to be issued.
NEW_DISP is the priority of an instruction which has its operands ready(depends on the spec_load_exec) and is ready to be issued.
CACHE_RELOAD on the other hand is the priority placed on an instruction which had previously had a cache miss and now the cache is ready so it wake up the instruction by setting its priority to CACHE_RELOAD. As the priority is higher it is selected to be issued.

core/LSU.hpp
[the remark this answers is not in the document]
knute-sifive: Ok, I can buy that for a dollar 💵 (sorry, bad movie reference).

core/LSU.hpp
h0lyalg0rithm: Yes I should rename it, its made sense when I had multiple queues

arupc:
> > Can you please explain what these priorities / states mean and when they would occur? Not sure I understand what is meant by 'mss ack' and 'another oustanding miss finish'
> > ```
> >                 HIGHEST = 0,
> >                 __FIRST = HIGHEST,
> >                 CACHE_RELOAD,   // Receive mss ack, waiting for cache re-access
> >                 CACHE_PENDING,  // Wait for another outstanding miss finish
> >                 MMU_RELOAD,     // Receive for mss ack, waiting for mmu re-access
> >                 MMU_PENDING,    // Wait for another outstanding miss finish
> >                 NEW_DISP,       // Wait for new issue
> >                 LOWEST,
> > ```
> > 
> > 
> > Does `NEW_DISP` simply mean dispatched and waiting to be issued?
> 
> The priority values are used to select the instruction that needs to be issued. NEW_DISP is the priority of an instruction which has its operands ready(depends on the spec_load_exec) and is ready to be issued. CACHE_RELOAD on the other hand is the priority placed on an instruction which had previously had a cache miss and now the cache is ready so it wake up the instruction by setting its priority to CACHE_RELOAD. As the priority is higher it is selected to be issued.

arupc:
> > > Can you please explain what these priorities / states mean and when they would occur? Not sure I understand what is meant by 'mss ack' and 'another oustanding miss finish'
> > > ```
> > >                 HIGHEST = 0,
> > >                 __FIRST = HIGHEST,
> > >                 CACHE_RELOAD,   // Receive mss ack, waiting for cache re-access
> > >                 CACHE_PENDING,  // Wait for another outstanding miss finish
> > >                 MMU_RELOAD,     // Receive for mss ack, waiting for mmu re-access
> > >                 MMU_PENDING,    // Wait for another outstanding miss finish
> > >                 NEW_DISP,       // Wait for new issue
> > >                 LOWEST,
> > > ```
> > > 
> > > 
> > > Does `NEW_DISP` simply mean dispatched and waiting to be issued?
> > 
> > 
> > The priority values are used to select the instruction that needs to be issued. NEW_DISP is the priority of an instruction which has its operands ready(depends on the spec_load_exec) and is ready to be issued. CACHE_RELOAD on the other hand is the priority placed on an instruction which had previously had a cache miss and now the cache is ready so it wake up the instruction by setting its priority to CACHE_RELOAD. As the priority is higher it is selected to be issued.

How would we implement an issue policy which prioritizes requests with MMU pending over request with CACHE pending?

How about instead of IssuePriority enum, we keep something equivalent to bitmasks, specifying for each instance of LoadStoreInstInfo, whether MMU_LOOKUP is needed and has been satisfied, CACHE_LOOKUP is needed and has been satisfied.  

This would enable implementation of various issue policies, without changing the enum. But this can be done as apart of another MR

core/LSU.cpp
```diff
@@ -188,6 +219,11 @@ namespace olympia
             // however, we can ONLY update instruction status as SCHEDULED for a new issue event
 
             ILOG("Another issue event scheduled " << inst_ptr);
+
+            if(isReadyToIssueInsts_())
```
arupc, line 223: should it be an assert instead of if(....), since we just updated issue priority? Does that mean we are guaranteed to find one inst that is ready?

core/LSU.cpp
```diff
@@ -200,61 +236,181 @@ namespace olympia
...
+            uev_issue_inst_.schedule(sparta::Clock::Cycle(0));
+        }
 
         ILOG("ROB Ack: Retired store instruction: " << inst_ptr);
     }
 
     // Issue/Re-issue ready instructions in the issue queue
     void LSU::issueInst_()
     {
+        if(allow_speculative_load_exec_ && replay_buffer_.size() > replay_buffer_size_){
```
arupc, line 251: why are we checking if allow_speculative_load_exec is true here? Even if speculative load exec is false, a load or a store can miss in MMU or cache and result in replay and hence, would allocate into the replay buffer during issue.

core/LSU.cpp
```diff
@@ -200,61 +236,181 @@ namespace olympia
...
         // Instruction issue arbitration
         const LoadStoreInstInfoPtr & win_ptr = arbitrateInstIssue_();
         // NOTE:
         // win_ptr should always point to an instruction ready to be issued
         // Otherwise assertion error should already be fired in arbitrateInstIssue_()
+        ILOG("Arbitrated inst " << win_ptr << " " << win_ptr->getInstPtr());
 
         lsu_insts_issued_++;
 
+        if(allow_speculative_load_exec_)
```
arupc, line 264: same comment as above

core/LSU.cpp
```diff
@@ -16,15 +17,25 @@ namespace olympia
...
+        replay_buffer_("replay_buffer", p->replay_buffer_size, getClock()),
+        replay_buffer_size_(p->replay_buffer_size),
+        replay_issue_delay_(p->replay_issue_delay),
         load_store_info_allocator_(sparta::notNull(OlympiaAllocators::getOlympiaAllocators(node))->
                                    load_store_info_allocator),
         memory_access_allocator_(sparta::notNull(OlympiaAllocators::getOlympiaAllocators(node))->
-                                 memory_access_allocator)
+                                 memory_access_allocator),
+        ldst_pipeline_("LoadStorePipeline",
+                       static_cast<uint32_t>(p->mmu_lookup_stage_length + p->cache_lookup_stage_length + p->cache_read_stage_length + 2), getClock()),
```
arupc, line 28: Can you please comment why we need +2?

core/LSU.hpp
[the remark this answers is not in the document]
h0lyalg0rithm: I made the change, Previously I was  changing them through a function call switch to constructor initialization

core/LSU.cpp
h0lyalg0rithm: Yes isReadytoIssueInsts_ finds a instruction that can be issued.Yes if it returns true it should issue.

core/LSU.cpp
h0lyalg0rithm: We use the replay buffer only when speculative_load_exec is true. I can remove the check since replay_buffer would default initialize to the size of the load store queue

core/LSU.cpp
h0lyalg0rithm: Its for the address_generation and completion stage length, currently they are of length 1 + 1

core/LSU.cpp
```diff
@@ -65,15 +79,24 @@ namespace olympia
         ldst_pipeline_.setContinuing(true);
 
         ldst_pipeline_.registerHandlerAtStage
-            (static_cast<uint32_t>(PipelineStage::MMU_LOOKUP),
+            (static_cast<uint32_t>(ADDRESS_CALCULATION),
```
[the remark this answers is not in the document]
h0lyalg0rithm: Yeah my bad should get rid of all them now

core/LSU.cpp
```diff
@@ -470,6 +880,19 @@ namespace olympia
         ILOG("Append new load/store instruction to issue queue!");
     }
 
+    void inline LSU::removeFromQueue_(LoadStoreIssueQueue &queue, const LoadStoreInstInfoPtr & inst_to_remove){
```
knute-sifive, line 883: Remove the `inline` keyword

core/LSU.cpp
```diff
@@ -482,16 +905,56 @@ namespace olympia
             }
         }
 
-        sparta_assert(false, "Attempt to complete instruction no longer exiting in issue queue!");
+        sparta_assert(false, "Attempt to complete instruction no longer exiting in issue queue! " << inst_ptr);
+    }
+
+    void inline LSU::appendToQueue_(LoadStoreIssueQueue &queue, const LoadStoreInstInfoPtr &inst_info_ptr){
```
knute-sifive, line 911: Remove the `inline` keyword

core/LSU.cpp
```diff
@@ -482,16 +905,56 @@ namespace olympia
...
+        sparta_assert(queue.size() < ldst_inst_queue_size_,
+                      "Appending load queue causes overflows!");
+
+        bool exists = std::find_if(
+            queue.begin(), queue.end(),
+            [inst_info_ptr](auto &queue_inst_ptr) { return queue_inst_ptr->getInstPtr() == inst_info_ptr->getInstPtr(); }
+        ) == std::end(queue);
+
+        sparta_assert(exists,  "Cannot push duplicate instructions into the queue " << inst_info_ptr);
```
knute-sifive, line 920: If you add a variable to the `LoadStoreInstInfo` that captures where it is, you can double check to make sure it's not already enqueued:
```
    sparta_assert(false == inst_info_ptr->isInIssueQueue(), "Instruction already in queue: " << inst_info_ptr);
    inst_info_ptr->setInIssueQueue(true);
```

This removes the expensive `find_if`

core/LSU.cpp
```diff
@@ -452,12 +682,192 @@ namespace olympia
...
+        }
+        if (isReadyToIssueInsts_())
+        {
+            ILOG("replay ready issue");
+            uev_issue_inst_.schedule(sparta::Clock::Cycle(0));
+        }
+    }
+
+    void LSU::updateInstReplayReady_(const InstPtr &inst_ptr){
+        for(auto iter = replay_buffer_.begin(); iter != replay_buffer_.end(); iter++)
```
knute-sifive, line 712: Why is the `InstPtr` being passed here?  Shouldn't it be a `LoadStoreInfo` type?  If you change this (strong suggestion you do), you can stash the replay iterator in the `LoadStoreInfo` and perform one operation:
```
replay_buffer_->erase(load_store_info_->getReplayIterator());
```

core/LSU.cpp
```diff
@@ -482,16 +905,56 @@ namespace olympia
...
+                if (!(*iter)->isReady())
+                {
+                    continue;
+                }
+                if (!(*win_ptr_iter)->isReady() || ((*iter)->winArb(*win_ptr_iter)))
+                {
+                    win_ptr_iter = iter;
+                }
+            }
+            if (win_ptr_iter.isValid() && (*win_ptr_iter)->isReady() && allow_speculative_load_exec_)
```
knute-sifive, line 947: Remove `allow_speculative_load_exec_` from the condition.  You've already checked for this.

h0lyalg0rithm:
Thanks just fixed the conflicts @klingaard

klingaard:
@h0lyalg0rithm can you provide an update on where you are with this PR?  Are you waiting on our approval now?

h0lyalg0rithm:
@klingaard Yes I am waiting on the approval

klingaard:
Ok, please resolve all conversations that you feel are addressed.  Thanks!

knute-sifive:
Did you test this with Kunal L2Cache stuff?  Does it work?

test/sim/CMakeLists.txt
```diff
@@ -12,54 +12,86 @@ file(CREATE_LINK ${SIM_BASE}/mavis/json ${CMAKE_CURRENT_BINARY_DIR}/mavis_isa_fi
...
+#  --report-yaml-replacements OUT_BASE coremark_test_report OUT_FORMAT text INST_START 0
+#  --report-search-dir reports
+#  --report reports/core_report.def
+#  --workload traces/core_riscv.zstf)
+#
+#sparta_named_test(olympia_json_test_missing_opcodes olympia
+#  --workload json_tests/missing_opcodes.json)
+#
+#sparta_named_test(olympia_json_test_fp_transfers olympia
+#  --workload json_tests/fp_transfers.json)
```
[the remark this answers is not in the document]
h0lyalg0rithm: It must be a mistake let me take a look

h0lyalg0rithm:
> Did you test this with Kunal L2Cache stuff? Does it work?

I havent been able to test it out, I tried to keep the PR in sync with master

h0lyalg0rithm:
@klingaard I rebuilt the project(cleant cache, cmake build folder) I ran both coremark and drhystone and both of them ran successfully.
I have the logs from valgrind but there were tons of errors but not sure if they are false positive

` ERROR SUMMARY: 10000034 errors from 41 contexts`

klingaard:
I'll take a look this evening.

core/LSU.cpp
```diff
@@ -216,22 +252,132 @@ namespace olympia
...
+        // calling pop below
+        if(allow_speculative_load_exec_)
+        {
+            ILOG("Appending to replay queue " << win_ptr);
+            appendToReplayQueue_(win_ptr);
+        }
+
+        // Remove inst from ready queue
+        ready_queue_.pop();
+        win_ptr->setInReadyQueue(false);
```
klingaard, line 272: This is the valgrind error I found.  After the `pop` operation, `win_ptr` is no longer valid (although the memory is still on the stack).  

What is happening:
1. You call `arbitrateInstIssue` which returns the Ptr by reference and you make a `const` reference to be used locally
2. Then, here, you `pop` it and now that pointer is nullified.

Suggestions on fixing:
Have `arbitrateInstIssue` return the pointer by value AND perform the `pop` _in that function_.

klingaard: Looks good.  After making another fix to DCache.hpp, valgrind is clean.  Good stuff!

knute-sifive:
I forgot one more thing.  😄  Can you run run clang format on the LSU since you're the last person to make major changes to it?

klingaard:
@h0lyalg0rithm, I hijacked your PR to update the names of the CI regressions (and merged with master).  FYI...

h0lyalg0rithm:
@klingaard I ran the formatter on most of the files I touched.I left some like `Inst.hpp` and `sim/Olympia.cpp`. 
I found an duplicate config in the clang format file. Could you please check I removed the correct one

.clang-format
```diff
@@ -138,7 +138,6 @@ ReflowComments:  true
 RemoveBracesLLVM: false
 SeparateDefinitionBlocks: Always
 ShortNamespaceLines: 1
-SortIncludes:    CaseSensitive
```
h0lyalg0rithm, line 141: The same key is present here https://github.com/riscv-software-src/riscv-perf-model/pull/92/commits/44291791647e5e562f4a1820be98555cb7f81668#diff-1026e0038b722990204a42bed8a6f7c0ec2302aa79e3fad1959d62ba968edfa2L192

klingaard:
Thanks for running the formatter.  One last, last thing, promise!  Can you update the Description field on what this PR did -- the changes it implemented and new expected behaviors.  That will be the final commit message.

core/LSU.hpp
```diff
@@ -26,29 +27,37 @@
...
-            }
+            LSUParameterSet(sparta::TreeNode* n) : sparta::ParameterSet(n) {}
 
             // Parameters for ldst_inst_queue
             PARAMETER(uint32_t, ldst_inst_queue_size, 8, "LSU ldst inst queue size")
+            PARAMETER(uint32_t, replay_buffer_size, ldst_inst_queue_size, "Replay buffer size")
+            PARAMETER(uint32_t, replay_issue_delay, 3, "Replay Issue delay")
+            // LSU microarchitecture parameters
+            PARAMETER(
+                bool, allow_speculative_load_exec, true,
```
klingaard, line 55: @h0lyalg0rithm is this parameter's boolean meaning inverted?   I'm looking at the code and if this is set to `true`, then the LS will NOT allow a load to bypass an older store.

h0lyalg0rithm: @klingaard This parameter by default is set to true, so it should allow a load to complete before a store

PR #102, opened 2023-10-16
Apply clang format

knute-sifive:
First cut of applied Clang format.  There's quite a lot, and I think we still need to tune this.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -107,9 +107,6 @@ IndentRequires:  false
 IndentWidth:     4
 IndentWrappedFunctionNames: false
 InsertTrailingCommas: None
-InsertBraces: true
-JavaScriptQuotes: Leave
-JavaScriptWrapImports: true
 KeepEmptyLinesAtTheStartOfBlocks: true
 LambdaBodyIndentation: Signature
 MacroBlockBegin: ''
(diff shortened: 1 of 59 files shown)


core/CPUTopology.cpp
knute-sifive, line 1: I don't like what clang-format did here.  Either we can figure out the magic command to _not_ do this or just take this file out of the list of files to format.

.clang-format
```diff
@@ -57,7 +57,7 @@ BreakConstructorInitializersBeforeComma: false
 BreakConstructorInitializers: BeforeColon
 BreakAfterJavaFieldAnnotations: false
 BreakStringLiterals: true
-ColumnLimit:     0
+ColumnLimit:     100
```
h0lyalg0rithm, line 60: I did not add this specifically as we would need to change all the files.
Since this PR includes the change :+1:

knute-sifive: Seems that all the files are touched regardless of this restriction.  😄   But might as well bite the bullet on it!

klingaard:
Reopened with _just the clang format rules_.  Will create a new PR with some files converted.

klingaard (approved):

PR #105, opened 2023-10-24
Jeffnye.dromajo stf update

jeffnye-gh:
This branch uses latest dromajo f3c3112, the changes move stf start/stop macro detection and trace generation into include/dromajo_template.h, within glue(riscv_cpu_interp, XLEN)(RISCVCPUState *s, int n_cycles) 

I have done limited testing with dhrystone and coremark, single cpu, with zstf and stf. 
stf_diff reports no differences. (clarify: no differences between this and stf's generated by previous impl.)

The main loop body can exit w/ a goto, illegal_instructions/etc. 
The tracing code makes the assumption that exceptions are not traced. 
I'm looking at this more. 

other minor changes:
- moved some globals into a struct, stf_writer remains global
- I've added a switch --no_stf_priv_check which can disable that check in the tracing enabled loop
- added tohost/fromhost  (clarify: struct variables added only, in advance of support)

### DIFF
--- a/traces/stf_trace_gen/dromajo_stf_lib.patch
+++ b/traces/stf_trace_gen/dromajo_stf_lib.patch
@@ -1,5 +1,5 @@
 diff --git a/CMakeLists.txt b/CMakeLists.txt
-index a3f932a..70c6cb3 100644
+index a3f932a..9c93d56 100644
 --- a/CMakeLists.txt
 +++ b/CMakeLists.txt
 @@ -39,7 +39,7 @@
@@ -11,6 +11,15 @@ index a3f932a..70c6cb3 100644
  project(dromajo)
  option(TRACEOS "TRACEOS" OFF)
  option(SIMPOINT "SIMPOINT" OFF)
+@@ -50,7 +50,7 @@ option(WARMUP "WARMUP" OFF)
+ 
+ add_compile_options(
+         -g
+-        -std=c++11
++        -std=c++17
+         -g
+         -Wall
+         -Wno-parentheses
 @@ -138,6 +138,22 @@ else ()
    target_link_libraries(dromajo_cosim_test dromajo_cosim)
  endif ()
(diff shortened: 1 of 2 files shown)


traces/stf_trace_gen/dromajo_stf_lib.patch
```diff
@@ -81,212 +213,107 @@ index 2f71c99..2de97a3 100644
...
++            "       --no_stf_priv_check Turn off the privledge check in STF generation\n"
+ #ifdef LIVECACHE
+             "       --live_cache_size live cache warmup for checkpoint (default 8M)\n"
+ #endif
+@@ -618,10 +620,12 @@ RISCVMachine *virt_machine_main(int argc, char **argv) {
      long        ncpus                    = 0;
      uint64_t    maxinsns                 = 0;
      uint64_t    trace                    = UINT64_MAX;
-+    const char *stf_trace                = nullptr;
++    const char *stf_trace                = NULL;
```
knute-sifive, line 271: Why change this to `NULL`?  `nullptr` is preferred simply for the fact it's a type.  `NULL` is just a `#define` which can be set to anything...

traces/stf_trace_gen/dromajo_stf_lib.patch
```diff
@@ -81,212 +213,107 @@ index 2f71c99..2de97a3 100644
...
-+
-     if (!do_trace) {
-         return keep_going;
-     }
-@@ -223,7 +322,7 @@ int main(int argc, char **argv) {
-     if (!m)
-         return 1;
- 
--    int n_cycles = 10000;
-+    int n_cycles = 1;
```
knute-sifive, line 230: Can this be changed back to 10k?

traces/stf_trace_gen/dromajo_stf_lib.patch
jeffnye-gh: i copied the style found in the original source.

e.g. const char *cmdline  = NULL; //as in dromajo_main.cpp

I will change it to nullptr

traces/stf_trace_gen/dromajo_stf_lib.patch
jeffnye-gh: yes it can
mistake in the patch, my source has 10K

klingaard:
Hey @jeffnye-gh, do you think you can finish this PR up and mark it for review?  This is good stuff.

jeffnye-gh:
Hi Knute,

Previous implementation was not as clean as it looked. 

I have an updated implementation on a branch in https://github.com/Condor-Performance-Modeling/dromajo, but I have not prepared a new PR. 

The remaining problem is a stack pointer (x2) initialization difference at startup in dhrystone on linux. I need to figure out why.

- 1:	0000000000010210 : 000066a2 c.ldsp      x13,8(x2) MEM 0000003fffa49ca8: [ 0000000000000000 ]
+ 1:	0000000000010210 : 000066a2 c.ldsp      x13,8(x2) MEM 0000003fffb53c98: [ 0000000000000000 ]

Day job priorities got in the way, I'll be able to get back to this on the weekend.

jeff

knute-sifive:
No worries.  Just wanted to make sure you weren't waiting on anything from our end.

PR #108, opened 2023-10-31
Raise exception in LSU when ROB is drained

h0lyalg0rithm:
Based on the discussion in #90


klingaard (commented):
Function name does not follow style guide

h0lyalg0rithm:
> Function name does not follow style guide
My bad just fixed it

core/LSU.cpp
```diff
@@ -92,6 +95,11 @@ namespace olympia
              << ": "
              << memory_access_allocator_.getNumAllocated()
              << " MemoryAccessInfo objects allocated/created");
+
+        if(retire_done_and_is_drained_){
+            bool ldst_queue_empty_ = ldst_inst_queue_.empty();
+            sparta_assert(ldst_queue_empty_, "Issue queue has pending instructions");
```
knute-sifive, line 101: Probably can combine and drop the local variable:
```suggestion
            sparta_assert(ldst_inst_queue_.empty(), "Issue queue has pending instructions");
```

test/core/rename/Rename_test.cpp
```diff
@@ -133,6 +133,14 @@ class olympia::LSUTester
         // while the ADD instruction is running, the STORE instruction should NOT issue
         EXPECT_TRUE(lsu.lsu_insts_issued_ == 0);
     }
+
+    void clear_entries(olympia::LSU &lsu){
```
knute-sifive, line 137: Granted this is a test, but the function name doesn't conform to standards

core/CPUFactory.cpp
```diff
@@ -94,6 +94,10 @@ auto olympia::CPUFactory::bindTree_(sparta::RootTreeNode* root_node,
                 setTLB(*private_nodes_.at(num_of_cores)->getResourceAs<olympia::SimpleTLB>());
         (core_tree_node->getChild("preloader")->getResourceAs<olympia::Preloader>())->
             preload();
+        auto rob = core_tree_node->getChild("rob")->getResourceAs<olympia::ROB>()->getContainer();
+        auto lsu = core_tree_node->getChild("lsu")->getResourceAs<olympia::LSU>();
+        rob->registerForNotification<bool, LSU, &LSU::onRobDrained_>
+            (lsu, "rob_notif_channel");
```
knute-sifive, line 100: Go ahead and use the MACRO for this:
```suggestion
        rob->REGISTER_FOR_NOTIFICATION(onRobDrained_, bool, "rob_notif_channel");
```

test/core/rename/Rename_test.cpp
h0lyalg0rithm: @knute-sifive  I wanted to know if this approach of fixing the test is better or should I add a toggle to turn the notification off/on.
Both the approaches require the next person using the simulator to remember to drain  the lsu if they run short simulations for their tests like in this case.

core/CPUFactory.cpp
h0lyalg0rithm: Sounds good

h0lyalg0rithm: @knute-sifive Looks like the macro needs to be defined in the class as it references (this)

test/core/rename/Rename_test.cpp
knute-sifive: You could send the notification like Retire does if you know the test is good to end.

knute-sifive: Or call the callback explicitly in the tester

h0lyalg0rithm: I think the call is more explicit, but now everytime some one will write a functional or end to end test they would end up getting the error because they used the LSU in their test

knute-sifive: I think that's ok.  If someone wants to include the LSU in their functional test, they need to indicate to the LSU that the they believe the test is done.  Otherwise, the LSU might be indicating to them an error ... which is a good thing.

core/ROB.cpp
```diff
@@ -198,6 +204,7 @@ namespace olympia
 
     void ROB::onStartingTeardown_() {
         if ((reorder_buffer_.size() > 0) && (false == rob_stopped_simulation_)) {
+            rob_drained_notif_source_->postNotification(true);
```
danbone, line 207: Based on the discussion, isn't this supposed to be set on the else branch?

> When Retire is completely drained AND is stopping simulation, that notification can be broadcasted, which LS can "hear" and set a boolean retire_done_and_is_drained_ and then you can check the ensure the LS issue queue is actually drained.

https://github.com/riscv-software-src/riscv-perf-model/pull/90#issuecomment-1739405558

knute-sifive: ☝🏻 good question (and good catch)

I think it's better if that notification is posted _at the point_ the ROB stops simulation on the scheduler.  And the LS should implement the `onStartingTeardown_` and check for this condition there and not the destructor.  That'd be pretty clean and ordering will be consistent.

core/CPUFactory.cpp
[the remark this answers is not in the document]
h0lyalg0rithm: I put it here so other components who would like to register for the same trigger are all grouped together over here.

core/ROB.cpp
h0lyalg0rithm: @knute-sifive I couldnt find find a good place to send the notification. The rob stops the scheduler only when ilimit parameter is passed to olympia but what about the case when it is stopped naturally

h0lyalg0rithm: > Based on the discussion, isn't this supposed to be set on the else branch?
> 
> > When Retire is completely drained AND is stopping simulation, that notification can be broadcasted, which LS can "hear" and set a boolean retire_done_and_is_drained_ and then you can check the ensure the LS issue queue is actually drained.
> 
> [#90 (comment)](https://github.com/riscv-software-src/riscv-perf-model/pull/90#issuecomment-1739405558)

Good point, I updated the code.

core/CPUTopology.cpp
```diff
@@ -68,6 +68,14 @@ olympia::CoreTopologySimple::CoreTopologySimple(){
             0,
             &factories->execute_rf
         },
+        {
```
h0lyalg0rithm, line 71: @knute-sifive I moved this above the lsu so it is destroyed before the lsu

core/CPUFactory.cpp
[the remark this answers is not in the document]
knute-sifive: I see and good forethought, but putting on OOP hat, the CPU factory shouldn't be responsible for ensuring the LSU is notified of the event.  Also, the code here requires the factory to be "aware" of the private function in the LSU.  I think it's best to move it there.

core/CPUTopology.cpp
knute-sifive: Hmmmm.... you shouldn't rely on that.  What if someone changes the order in the future?  What if the framework changes the order/destruction order (there are NO rules in Sparta that dictate that).  Future developers might get indeterministic behavior that they will have to chase down.  Best to ensure the notification is sent before any teardown occurs.

core/CPUFactory.cpp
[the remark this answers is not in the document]
h0lyalg0rithm: I moved the setup to the lsu

core/ROB.cpp
```diff
@@ -188,6 +194,8 @@ namespace olympia
     {
         if(getClock()->currentCycle() - last_retirement_ >= retire_timeout_interval_)
         {
+            if(reorder_buffer_.empty())
```
h0lyalg0rithm, line 197: @knute-sifive I had to add this else the event is scheduled forever and simulation doesnt end for some of the tests with small input like example.json file

knute-sifive:
I pushed changes to this PR to origin/lsu_drained.  Please pull those changes into your fork.

h0lyalg0rithm:
> I pushed changes to this PR to origin/lsu_drained. Please pull those changes into your fork.

Thanks Knute

klingaard:
At your earliest convenience, @h0lyalg0rithm can you update this branch (resolve conflicts)?

h0lyalg0rithm:
@klingaard I synced it up with master

klingaard (approved):

PR #110, opened 2023-10-31
Use latest map which includes priority queue

h0lyalg0rithm:


.github/workflows/ubuntu-build.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         with:
           repository: sparcians/map
           path: map
-          ref: map_v2
+          ref: 16bc957d28bd1c3cd7038ade910774fa4f1e17c3
```
knute-sifive, line 39: You can change the ref to map_v2.0.10

klingaard (approved):

PR #111, opened 2023-11-03
Change SimpleDL1 to CacheFuncModel

vineethkm:
Change name of the SimpleDL1 class to a more generic name "SimpleCache3"

### DIFF
--- a/core/CacheFuncModel.hpp
+++ b/core/CacheFuncModel.hpp
@@ -87,15 +87,15 @@ namespace olympia
 
     }; // class SimpleCacheLine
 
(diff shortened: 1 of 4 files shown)


klingaard:
Why the number 3?

vineethkm:
> Why the number 3?

SimpleDL1 inherits from SimpleCache2. I could rename it to something along the lines of "CacheFuncModel" instead as suggested by Kunal

knute-sifive:
> I could rename it to something along the lines of "CacheFuncModel" instead as suggested by Kunal

Yeah, let's do that.

knute-sifive (approved):

PR #114, opened 2023-11-07
Adding L2Cache Model - new sparta Unit

Kunal-Buch:
-> L2Cache can accept multiple requests from DCache and IL1.
-> Cache functional model used is SimpleDL1.
-> Configurable cache hit latency. 
-> Misses are forwarded to BIU and kept are tracked in the local buffer.
-> Miss to the pending cacheline is not sent out to BIU.
-> Backpressure handling for BIU credits.
-> make regress clean.


core/DCache.cpp
```diff
@@ -78,17 +81,21 @@ namespace olympia {
...
+    void DCache::getRespFromL2Cache_(const InstPtr &inst_ptr) {
         out_lsu_lookup_req_.send(cache_pending_inst_);
         reloadCache_(inst_ptr->getRAdr());
         cache_pending_inst_.reset();
         busy_ = false;
     }
 
+    void DCache::getAckFromL2Cache_(const bool &ack) {
+        // Process ACK
+    }
```
klingaard, line 99: What do expect the DCache to do with the acknodelgement?  (in the meantime consider adding a `(void) ack;` to prevent some compilers from complaining about unused parameter).

klingaard (commented):
About half way through...  😄

core/LSU.hpp
```diff
@@ -399,6 +399,30 @@ namespace olympia
 
     };
 
+    inline std::ostream & operator<<(std::ostream & os,
+        const olympia::MemoryAccessInfo::UnitName & unit) {
```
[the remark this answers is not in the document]
Kunal-Buch: Done. Also corrected it for the other mis-indented functions. :)

core/LSU.hpp
```diff
@@ -399,6 +399,30 @@ namespace olympia
...
+                os << "DCACHE";
+                break;
+            case olympia::MemoryAccessInfo::UnitName::L2CACHE:
+                os << "L2CACHE";
+                break;
+            case olympia::MemoryAccessInfo::UnitName::BIU:
+                os << "BIU";
+                break;
+            default:
+                throw sparta::SpartaException("NUM_STATES cannot be a valid enum state.");
```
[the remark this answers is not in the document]
Kunal-Buch: That's a good idea!

mss/L2Cache.hpp
```diff
@@ -0,0 +1,374 @@
...
+            uint32_t __FIRST = NO_ACCESS;
+            uint32_t CACHE_LOOKUP;
+            uint32_t HIT_MISS_HANDLING;
+            uint32_t NUM_STAGES;
+            uint32_t __LAST;
+        };
+        
+	    // Skipping the use of SpartaSharedPointer due to allocator bug
+	    // Instead using std::shared_ptr. That works cleanly.
+        using L2MemoryAccessInfoPtr = std::shared_ptr<olympia::MemoryAccessInfo>;
```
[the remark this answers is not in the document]
Kunal-Buch: This is a remnant from the older olympia fork with which, the deallocation of SpartaSharedPointer ref_count was not happening cleanly. But std::shared_ptr worked fine. 

Maybe I can now give it a try again and see if it works.

mss/L2Cache.hpp
```diff
@@ -0,0 +1,374 @@
...
+        sparta::Clock::Cycle pipeline_req_issue_delay_ = 0;
+        
+        sparta::Buffer<std::shared_ptr<olympia::MemoryAccessInfo>> miss_pending_buffer_;
+        const uint32_t miss_pending_buffer_size_;
+        
+
+        // L2 Cache
+        using CacheHandle = olympia::SimpleDL1::Handle;
+        uint32_t l2_lineSize_;
+        uint8_t shiftBy_;
```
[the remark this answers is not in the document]
Kunal-Buch: Yes. Thank you.. :)

core/MemoryAccessInfo.hpp
```diff
@@ -33,12 +33,25 @@ namespace olympia {
...
             NO_ACCESS = 0,
             __FIRST = NO_ACCESS,
+            RELOAD,
             MISS,
             HIT,
             NUM_STATES,
             __LAST = NUM_STATES
         };
 
+        enum class UnitName : std::uint32_t {
```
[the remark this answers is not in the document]
Kunal-Buch: How about 'ArchUnit' ?

core/DCache.hpp
```diff
@@ -51,8 +53,11 @@ namespace olympia {
         sparta::DataInPort<MemoryAccessInfoPtr> in_lsu_lookup_req_
                 {&unit_port_set_, "in_lsu_lookup_req", 0};
 
-        sparta::DataInPort<InstPtr> in_biu_ack_
-                {&unit_port_set_, "in_biu_ack", 1};
+        sparta::DataInPort<bool> in_l2cache_ack_
+                {&unit_port_set_, "in_l2cache_ack", 1};
```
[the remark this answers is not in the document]
Kunal-Buch: Thanks Knute! :) The Sparta::Signal is actually new to me. Should I change it in this PR or is it just a guidance to keep in mind for future development?

core/MemoryAccessInfo.hpp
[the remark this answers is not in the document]
klingaard: I like it!

core/DCache.hpp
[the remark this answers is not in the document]
klingaard: Give it a try in this PR!  Be interesting to see if you have any issues with it.

core/LSU.hpp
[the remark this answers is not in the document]
klingaard: Thanks!

mss/L2Cache.hpp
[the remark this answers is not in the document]
klingaard: Please do.  If there's a bug in SSP, I need to fix it!

core/MemoryAccessInfo.hpp
```diff
@@ -100,6 +121,10 @@ namespace olympia {
         // DCache access status
         CacheState cache_access_state_;
 
+        // Src and destination unit name for the packet
+        UnitName src_;
+        UnitName dest_;
```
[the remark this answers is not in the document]
klingaard: Thanks for doing that!  One thing (for the future or right now, up to you since it's a bit legacy) it's better, IMHO, to initialize these inline:
```suggestion
        UnitName src_ = UnitName::NO_ACCESS;
        UnitName dest_ = UnitName::NO_ACCESS;
```
Reason (not that've run into this issue before or anything 😛 ): if someone adds a new constructor and doesn't take advantage of construction delegation, they _will_ forget to initialize these variables.  I guarantee it!

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+            ev_handle_dcache_l2cache_req_.schedule(sparta::Clock::Cycle(0));
+            num_reqs_from_dcache_++;
+        }
+        else {
+            ILOG("This request cannot be serviced right now, L2Cache input buffer from DCache is already full!");
+        }
+    }
+
+    // Receive new L2Cache request from IL1
+    void L2Cache::getReqFromIL1_(const olympia::InstPtr & inst_ptr) {
```
klingaard, line 150: Just a nit: one function is named `getReqFromDCache_` and this one is called `getReqFromIL1_`.  Maybe a rename this to `getReqFromICache_`?  Feel free to ignore this request if you want...

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+        
+        ILOG("Request received from IL1 on the port");
+
+        appendIL1ReqQueue_(inst_ptr);
+
+        // Schedule L2Cache request handling event only when:
+        // (1)Request queue is not empty
+        if (il1_req_queue_.size() < il1_req_queue_size_) {
+            ev_handle_il1_l2cache_req_.schedule(sparta::Clock::Cycle(0));
+            num_reqs_from_il1_++;
```
klingaard, line 160: Another small nit:
```suggestion
            ++num_reqs_from_il1_;
```
The original code will return a _copy_ of the counter after the increment, which (mostly in debug) this is inefficient.  Most of the times the compiler will optimize this out, but don't rely on that.  😉

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+
+        // Schedule L2Cache request handling event only when:
+        // Request queue is not empty
+        if (dcache_req_queue_.size() < dcache_req_queue_size_) {
+            ev_handle_dcache_l2cache_req_.schedule(sparta::Clock::Cycle(0));
+            num_reqs_from_dcache_++;
+        }
+        else {
+            ILOG("This request cannot be serviced right now, L2Cache input buffer from DCache is already full!");
+        }
```
klingaard, line 146: Instead of scheduling `ev_handle_dcache_l2cache_req_`, shouldn't you just call `handle_DCache_L2Cache_Req_` directly?

Kunal-Buch: I derived this paradigm from BIU, and also thought scheduling an event creates a placeholder for when something fancier needs to be done, timing wise.

core/MemoryAccessInfo.hpp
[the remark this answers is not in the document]
Kunal-Buch: Done!

mss/L2Cache.cpp
Kunal-Buch: Done!

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+            reqPtr->setDestUnit(L2ArchUnit::DCACHE);
+
+            pipeline_req_queue_.push(reqPtr);
+            ILOG("DCache request is sent to Pipeline_req_Q!");
+
+            dcache_req_queue_.erase(dcache_req_queue_.begin());
+            
+            // Send out the ack to DCache for credit management
+            ev_handle_l2cache_dcache_ack_.schedule(sparta::Clock::Cycle(0));
+        }
```
knute-sifive, line 349: Might want an `else` clause here that just asserts...

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+        // Try to issue a request to l2cache_pipeline_
+        ev_issue_req_.schedule(sparta::Clock::Cycle(0));
+
+        // Schedule a ev_create_req_ event again to see if the the new request 
+        // from any of the requestors can be put into pipeline_req_queue_
+        if (   !biu_resp_queue_.empty()
+            || !icache_req_queue_.empty()
+            || !dcache_req_queue_.empty() )
+            
+            ev_create_req_.schedule(sparta::Clock::Cycle(1));
```
knute-sifive, line 360: Our coding style guide rule 17 requires that all conditional statements, even if just one line, must be in `{}`.  Reason for this:  I personally lost _years_ (I exaggerate 😆 ) of time debugging code like this:
```
if (condition) x = 3;
    y = 4;
```
WHY IS y always getting 4??!?  Oh.  Missing `{}`.

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+        if (hasCreditsForPipelineIssue_() && !pipeline_req_queue_.empty()) {
+            l2cache_pipeline_.append(pipeline_req_queue_.front());
+            ++inFlight_reqs_;
+            ILOG("Request is sent to Pipeline! SrcUnit : " << pipeline_req_queue_.front()->getSrcUnit());
+
+            pipeline_req_queue_.pop(); 
+        }
+
+        // Checking for the queue empty again before scheduling the event for the next clock cycle
+        if (!pipeline_req_queue_.empty())
```
knute-sifive, line 376: Another missing `{}` -- I'll let you hunt the other down... :wink:

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+            req->setCacheState(cacheLookUpResult);
+        }
+    }
+
+    // Pipeline Stage HIT_MISS_HANDLING
+    void L2Cache::handleCacheAccessResult_() {
+        const auto req = l2cache_pipeline_[stages_.HIT_MISS_HANDLING];
+        ILOG("Pipeline stage HIT_MISS_HANDLING : " << req->getInstPtr());
+        
+        inFlight_reqs_--;
```
knute-sifive, line 412: Suggest pre-decrement to avoid the copy.  Good habit to get into.

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+
+            // Handle the miss instruction by storing it aside while waiting 
+            // for lower level memory to return
+            if (miss_pending_buffer_.size() < miss_pending_buffer_size_) {
+
+                ILOG("Storing the CACHE MISS in miss_pending_buffer_");
+                miss_pending_buffer_.push_back(req);
+            }
+            else {
+                sparta_assert("No space in miss_pending_buffer_! Why did the frontend issue push the request onto l2cache_pipeline_?");
```
knute-sifive, line 438: Good assert!

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+            }
+
+            // Function to check if the request to the given cacheline is present in the miss_pending_buffer_ 
+            auto getCacheLine = [this] (auto reqPtr) { return reqPtr->getInstPtr()->getRAdr() >> shiftBy_; };
+            auto is_addr_present = [req, getCacheLine] (auto reqPtr)
+                            { return (req != reqPtr && getCacheLine(reqPtr) == getCacheLine(req));};
```
knute-sifive, line 444: To save a little performance why not create a cache (HA!) of the `req`'s cache line?  Also, pass the `reqPtr` as `const &` to avoid copy:
```suggestion
            auto getCacheLine = [this] (auto reqPtr) { return reqPtr->getInstPtr()->getRAdr() >> shiftBy_; };
            const auto req_cl = getCacheLine(req);
            auto is_addr_present = [&req, req_cl, getCacheLine] (const auto & reqPtr)
                            { return (req != reqPtr && getCacheLine(reqPtr) == req_cl); };
```

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+
+            // Send out the request to BIU for a cache MISS if it is not sent out already
+            auto reqIter = std::find_if(miss_pending_buffer_.begin(), miss_pending_buffer_.end(), is_addr_present);
```
knute-sifive, line 447: Curious, if it does hit, does it make sense to append this request to that miss so you can track how many requests are to that existing miss?  Kinda handy to know...

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+        sparta_assert(biu_req_queue_.size() <= biu_req_queue_size_ ,"BIU req queue overflows!");
+        
+        // Push new request to the biu_req_queue_ if biu credits are available with the L2Cache
+        if (biu_req_queue_.size() < biu_req_queue_size_) {
+            biu_req_queue_.emplace_back(inst_ptr);
+
+            ev_handle_l2cache_biu_req_.schedule(sparta::Clock::Cycle(0));
+        }
+        else {
+            sparta_assert("No space in biu_req_queue_! Why did the frontend issue push the request onto l2cache_pipeline_?");
```
knute-sifive, line 521: I didn't check all of the calls to `sparta_assert`, but your usage here is incorrect.  A string is _always_ true, so the assert will never fire.  Kinda surprised the compiler didn't pick up on that.  Suggested fixes:
```suggestion
            sparta_assert(false, "No space in biu_req_queue_! Why did the frontend issue push the request onto l2cache_pipeline_?");
            sparta_assert(!"No space in biu_req_queue_! Why did the frontend issue push the request onto l2cache_pipeline_?");

```

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+            else if (channel_select_ == Channel::DCACHE) {
+                channel_select_ = Channel::ICACHE;
+                if (!dcache_req_queue_.empty()) {
+                    winner = Channel::DCACHE;
+                    ILOG("Arbitration winner - DCache");
+
+                    return winner;
+                }
+            }
+        }
```
knute-sifive, line 599: Why is this in a `while(true)` loop?   There seems to be no iterations here -- and what happens if the `channel_select_` isn't one of I/Dcache?

mss/L2Cache.cpp
```diff
@@ -0,0 +1,657 @@
...
+    }
```
knute-sifive, line 600: This function _technically_ doesn't return anything.  `gcc` usually complains bitterly about that.  Two ways to solve:
1. Add an assert and return NO_ACCESS (which will never be reached)
2. Add `__builtin_unreachable();` if you think there is no path to the end of the function.

test/core/l2cache/L2SourceUnit.hpp
knute-sifive, line 1: This is really great!  Wonder if you could set up the source unit (via parameters) to expect specific number of requests, transactions, etc and assert if it doesn't see the right number of acks, etc.

core/DCache.cpp
Kunal-Buch: Ack Implementation: When DCache sends the request to L2Cache for a miss, this boolean will be set to false, and Dcache should wait for ack from L2Cache, notifying DCache, that there is "**_space in its dcache request buffer_**".
 
Set `dcache_l2cache_credit_available_` to true so that the following misses from DCache can be sent out to L2Cache.

mss/L2Cache.cpp
Kunal-Buch: You mean with a separate member of the request that counts the number of misses to the same/originally missed cacheline?

mss/L2Cache.cpp
Kunal-Buch: Hmm! I thought it was always false by default. I picked it up from usage elsewhere in the Olympia model.
But good practice to explicitly mention it anyway! :)

test/core/l2cache/L2SourceUnit.hpp
Kunal-Buch: Thank you! :)

So, it does that in some way in the destructor where it checks for pending_reqs_ and pending_acks_ to be 0 at the end of the test. These are dynamic counters incremented and decremented as the transactions flow through the L2SouceUnit.

This way, it is making sure that if the request was picked up from the .json file and L2Cache did not return and ack or response for it, then the test asserts. It has helped me debug many times already! :)

mss/L2Cache.cpp
knute-sifive: If you find it in other places in Olympia, please flag them -- that's a bug.  😉

mss/L2Cache.cpp
knute-sifive: Something like that.  You can "append" this request to the found request for reference or for debug.

core/CPUTopology.cpp
```diff
@@ -202,12 +210,28 @@ olympia::CoreTopologySimple::CoreTopologySimple(){
             "cpu.core*.lsu.ports.in_cache_free_req"
         },
         {
-            "cpu.core*.dcache.ports.out_biu_req",
+            "cpu.core*.dcache.ports.out_l2cache_req",
```
arupc, line 213: Can we add a `L2Cache.md` file documenting modeled microarchitcture of L2Cache? Such a document would help in understanding the code, and would enable new users and higher adoption by the community. 

For example, it can include:
* a description of various ports of L2Cache and their intended purpose
* various structures and pipelines used in L2cache
* stage by stage behavior of each pipeline

core/DCache.hpp
```diff
@@ -23,6 +23,7 @@ namespace olympia {
             PARAMETER(uint32_t, l1_associativity, 8, "DL1 associativity (power of 2)")
             PARAMETER(uint32_t, cache_latency, 1, "Assumed latency of the memory system")
             PARAMETER(bool, l1_always_hit, false, "DL1 will always hit")
+            PARAMETER(bool, l1_l2cache_credit_available, true, "Aloowing L1 cache to send the first miss request to L2Cache")
```
arupc, line 26: typo for `Allowing`

arupc: What happens if this is false? When will Dcache be able to send miss request to L2Cache?

core/MemoryAccessInfo.hpp
```diff
@@ -48,7 +61,9 @@ namespace olympia {
             mmu_access_state_(MMUState::NO_ACCESS),
 
             // Construct the State object here
-            cache_access_state_(CacheState::NO_ACCESS) {}
+            cache_access_state_(CacheState::NO_ACCESS),
+            src_(ArchUnit::NO_ACCESS),
+            dest_(ArchUnit::NO_ACCESS) {}
```
arupc, line 66: Can you please explain the semantics of src_ and dest_ of MemoryAccessInfo?

mss/L2Cache.cpp
Kunal-Buch: I think this was a remnant of the previous implementation. I have fixed this in my local change. Will push it soon. Thanks for pointing it out. :)

mss/L2Cache.cpp
Kunal-Buch: Took care of this as well along with the change for the comment above.

core/DCache.hpp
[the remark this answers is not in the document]
Kunal-Buch: I actually decided to change the credit protocol where the *ack_ port now is of type uint32_t, returning credits based on empty slots available in the input request buffers for each requestor.

core/DCache.cpp
Kunal-Buch: L2Cache _now_ returns 'n' credits based on the number of empty slots available in the requestor port input buffers. DCache should use this number to forward the requests to L2Cache as L2Cache will be able to hold that many number of pending requests from DCache.

*ack_ is now of type uint32_t.

core/CPUTopology.cpp
Kunal-Buch: Sure Arup.

Where should this L2Cache.md file reside?
I already have the info available in the presentation I prepared. Just need to port it over! :)

core/DCache.hpp
Kunal-Buch: This changed to a uint32_t type now. It will be programmed by L2Cache into DCache through the *ack_ port at the beginning of the simulation. This represents the number of l2cache_credits for DCache.

At every de-allocation from the input_buffer, L2Cache will update DCache with new number of empty slots in the input_buffer.

core/MemoryAccessInfo.hpp
Kunal-Buch: I added these members to MemoryAccessInfo for keeping track of the flow of transaction.
When the request arrives in L2Cache, I create a MemoryAccessInfo packet where I set the `src_ = dest_ = requestor`.

- If it hits in L2, the packet is returned to` dest_`.
- If it misses, the `dest_ = BIU`.
- When BIU returns, `src_ = BIU` and `dest_` becomes the original src where it needs to be returned to.

I am using the `src_` and `dest_` to determine where the packet is headed after the cacheLookup for HIT, MISS or RELOAD.

One major reason to have something like this is that L2Cache model supports requests from both _ICache_ and _DCache_. This aids in keeping track of the original requestor.

mss/L2Cache.hpp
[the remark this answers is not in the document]
Kunal-Buch: Changed shared_ptr to SSP. Not seeing the issue I was seeing earlier.

mss/L2Cache.cpp
Kunal-Buch: Implemented this. Each request in this buffer points to the next miss to the same cacheline through a member variable `next_req_`. By default/if there is no miss to the same cacheline after the said req in the buffer, `next_req_ = nullptr`.

Kunal-Buch:
Hello @arupc , @klingaard ,:

I have reviewed all the comments for the pull request above and provided the updates for all of them through the commits listed here.

Is there anything else that remains to be reviewed/updated?
Is it good to be approved for merge? :)

mss/L2Cache.hpp
```diff
@@ -0,0 +1,376 @@
...
+        
+        using L2ArchUnit = olympia::MemoryAccessInfo::ArchUnit;
+        using L2CacheState = olympia::MemoryAccessInfo::CacheState;
+        using L2CachePipeline = sparta::Pipeline<olympia::MemoryAccessInfoPtr>;
+        
+        PipelineStages stages_;
+        L2CachePipeline l2cache_pipeline_;
+        
+        sparta::Queue<olympia::MemoryAccessInfoPtr> pipeline_req_queue_;
+        const uint32_t pipeline_req_queue_size_;
```
klingaard, line 201: Is this used?  If not, feel free to delete it.  You can query the size from the `sparta::Queue`

klingaard (approved):
This looks good!  Thank you for addressing the comments posted.

klingaard:
@h0lyalg0rithm you might need to see how this impacts your current PR

Issue #116, opened 2023-11-08
Rename reference counters not decremented for AMO instructions


danbone:
There is a bug in the rename implementation which doesn't decrement the reference counter for RS2 mappings on AMO instructions, leading to a lock up once all rename registers are exhausted.

You can easily reproduce by using a trace of repeated AMOs like:

```
   {
        "mnemonic": "amoadd.w",
        "rs1": 1,
        "rs2": 1,
        "rd": 1
    },
   {
        "mnemonic": "amoadd.w",
        "rs1": 1,
        "rs2": 1,
        "rd": 1
    }
...
```

`olympia -p top.cpu.core0.rename.params.num_integer_renames 48 -l "top.cpu*" info info.log traces/amo_lockup.json `

No error message is printed for the lock up but you can see from the logfile that the trace hasn't finished:
```
{0000000084 00000084 top.cpu.core0.dispatch info} receiveCredits_: lsu got 1 credits, total: 8
{0000000084 00000084 top.cpu.core0.dispatch info} scheduleDispatchSession: no rob credits or no instructions to process
{0000000084 00000084 top.cpu.core0.lsu info} completeInst_: Complete Load Instruction: amoadd.w uid(16)
{0000000085 00000085 top.cpu.core0.rob info} retireInstructions_: num to retire: 1
{0000000085 00000085 top.cpu.core0.rob info} retireInstructions_: retiring uid: 16    RETIRED 0 pid: 16 'amoadd.w       1,1,1' 
{0000000085 00000085 top.cpu.core0.dispatch info} scheduleDispatchSession: no rob credits or no instructions to process
{0000000085 00000085 top.cpu.core0.dispatch info} robCredits_: ROB got 1 credits, total: 30
{0000000086 00000086 top.cpu.core0.rename info} getAckFromROB_: Retired instruction: uid: 16    RETIRED 0 pid: 16 'amoadd.w     1,1,1' 
{0000000086 00000086 top.cpu.core0.rename info} scheduleRenaming_: current stall: NO_RENAMES
{0000000000 -------- top.cpu.core0.rob info} ~ROB: ROB is destructing now, but you can still see this message
```

The bug appears to be caused due to a mismatch in conditions when updating the reference counter. The increment condition is based on `isLoadStoreInst() && exists(RS2)`, whereas the decrement condition is just `isStoreInst()`. Atomics only trigger the increment logic.

[...]

PR #117, opened 2023-11-08
Bug fix, tests, and assertion for stall condition

AaronGChan:
Fixing https://github.com/riscv-software-src/riscv-perf-model/issues/116

### DIFF
--- a/core/Rename.cpp
+++ b/core/Rename.cpp
@@ -358,10 +358,4 @@ namespace olympia
             ev_schedule_rename_.schedule(1);
         }
     }
-
-
-    void Rename::onStartingTeardown_(){
-        sparta_assert(current_stall_ == StallReason::NOT_STALLED, "Rename is stalled when entering teardown, possible lockup in Rename?")
-    }
-
 }
--- a/core/Rename.cpp
+++ b/core/Rename.cpp
@@ -132,18 +132,21 @@ namespace olympia
         }
 
         const auto & srcs = inst_ptr->getRenameData().getSourceList();
-        // decrement reference for store instruction to data register
-        if(inst_ptr->isStoreInst()) {
+        // decrement reference to data register
+        if(inst_ptr->isLoadStoreInst()){
             const auto & data_reg = inst_ptr->getRenameData().getDataReg();
-            --reference_counter_[data_reg.rf][data_reg.val];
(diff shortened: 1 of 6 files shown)


core/Rename.cpp
```diff
@@ -356,4 +359,9 @@ namespace olympia
         }
     }
 
+
+    void Rename::onStartingTeardown_(){
+        sparta_assert(current_stall_ == StallReason::NOT_STALLED, "Rename is stalled when entering teardown, possible lockup in Rename?")
```
AaronGChan, line 364: @knute-sifive Was thinking either to have an assert like this to ensure that there is no ending stall when teardown begins, or should I do a notification similar to https://github.com/riscv-software-src/riscv-perf-model/pull/108

knute-sifive: Yeah, you can't have this assertion.  It's possible that Rename is stalled waiting for free registers AND the ROB was told to stop simulation on instruction _N_.  This is a valid case and you don't want to assert on that. 

In #108 (still under review at this time), having the ROB broadcast "I stopped simulation" and Rename listening for that and determining if it's a cause in this function is appropriate.

core/Rename.cpp
```diff
@@ -132,18 +132,21 @@ namespace olympia
         }
 
         const auto & srcs = inst_ptr->getRenameData().getSourceList();
-        // decrement reference for store instruction to data register
-        if(inst_ptr->isStoreInst()) {
+        // decrement reference to data register
+        if(inst_ptr->isLoadStoreInst()){
```
knute-sifive, line 136: @danbone's fix is a good catch: noticing that the renaming and reclamation are not in sync

But there's something nagging me about all of this -- we're renaming `x0`.  Can you open a new issue that fixes that?

Specifically, in `renameInstructions_`, when iterating over the sources and destinations, you need to _skip_ renaming `x0`:  

```
if (num == 0 && rf == core_types::RF_INTEGER) { continue; }
```

knute-sifive (approved):
I'll approve this for now, but please put in a proper fix for the end-of-simulation assertion (use the ROB notification).

core/Rename.cpp
AaronGChan: I'll remove the teardown and assertion for now, and I'll add in the rename debugging once https://github.com/riscv-software-src/riscv-perf-model/pull/108 is merged.

Issue #118, opened 2023-11-08
Rename: Ignore renaming of x0


AaronGChan:
Current implementation will rename x0 and map it to a PRF, we should just be reading directly from x0.

knute-sifive:
Is this done?

AaronGChan:
> Is this done?

Yeah it is, just moved the PR to open from draft.

PR #120, opened 2023-11-14
Ignore Renaming of x0

AaronGChan:
Fix renaming to skip renaming of x0 outlined in https://github.com/riscv-software-src/riscv-perf-model/issues/118.

Changes:
- `renameInstructions_()` now checks if a source is x0, if it is and is a data operand for a LSU operation, we still create a data register for it, as we need to check in LSU for x0 before issuing instruction
- Added a boolean `x0` field to `Reg` struct in `RenameData` that tracks if a data register is an `x0` instruction. This helps on reclamation of rename credits in `getAckFromROB_()` to make sure we don't decrement for an x0 data register
- Updated logic to handle for destination reclamation in `getAckFromROB_()` to ensure when we reclaim credits/freelist registers/references, we don't process them for x0 destinations
- Fixed initial allocation of freelist registers for `RF_INTEGER` to only map registers for x1 - x31, before we were mapping from x0 - x31, so we should technically see a performance improvement as we gained 1 extra freelist register from not renaming x0

### DIFF
--- a/core/Inst.hpp
+++ b/core/Inst.hpp
@@ -40,7 +40,7 @@ namespace olympia
                 uint32_t val = 0;
                 core_types::RegFile rf = core_types::RegFile::RF_INVALID;
                 mavis::InstMetaData::OperandFieldID field_id = mavis::InstMetaData::OperandFieldID::NONE;
-                bool x0 = false;
+                bool is_x0 = false;
             };
             using RegList = std::vector<Reg>;
 
--- a/core/Inst.hpp
+++ b/core/Inst.hpp
@@ -40,6 +40,7 @@ namespace olympia
                 uint32_t val = 0;
                 core_types::RegFile rf = core_types::RegFile::RF_INVALID;
                 mavis::InstMetaData::OperandFieldID field_id = mavis::InstMetaData::OperandFieldID::NONE;
+                bool x0 = false;
             };
             using RegList = std::vector<Reg>;
 
--- a/core/Rename.cpp
+++ b/core/Rename.cpp
@@ -40,7 +40,7 @@ namespace olympia
                 (CREATE_SPARTA_HANDLER_WITH_DATA(Rename, getAckFromROB_, InstPtr));
         sparta::StartupEvent(node, CREATE_SPARTA_HANDLER(Rename, setupRename_));
         auto setup_map = [this] (core_types::RegFile reg_file, const uint32_t num_renames) {
-                        uint32_t num_regs = 32;  // default risc-v ARF count
+                        const uint32_t num_regs = 32;  // default risc-v ARF count
                         sparta_assert(num_regs < num_renames); // ensure we have more renames than 32 because first 32 renames are allocated at the beginning
                         // initialize the first 32 Float (31 for INT) regs, i.e x1 -> PRF1
 
@@ -101,9 +101,13 @@ namespace olympia
 
             // Initialize 32 scoreboard resources,
             // all ready.
+            uint32_t reg = 0;
+            if(rf == core_types::RegFile::RF_INTEGER){
+                reg = 1;
+            }
             constexpr uint32_t num_regs = 32;
             core_types::RegisterBitMask bits;
-            for(uint32_t reg = 0; reg < num_regs; ++reg) {
+            for(; reg < num_regs; ++reg) {
                 bits.set(reg);
             }
             scoreboards_[rf]->set(bits);
@@ -133,7 +137,8 @@ namespace olympia
             const auto dest = dests[0];
             const auto rf  = olympia::coreutils::determineRegisterFile(dest);
             const auto num = dest.field_value;
-            if (num != 0 || rf != core_types::RF_INTEGER)
+            const bool is_x0 = num != 0 || rf != core_types::RF_INTEGER;
+            if (is_x0)
             {
                 auto const & original_dest = inst_ptr->getRenameData().getOriginalDestination();
                 --reference_counter_[original_dest.rf][original_dest.val];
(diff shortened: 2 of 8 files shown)


danbone:
Missing logic here:
https://github.com/riscv-software-src/riscv-perf-model/blob/55661600abe6604d3fc369da398e7677efdb1b1c/core/Rename.cpp#L197-L201

danbone:
This change seems a bit convoluted, would it not be better to refactor the allocation and freeing into separate functions, and act differently when the input register is X0. 

Also, I can't find out where the scoreboard will always ready x0 reads?
https://github.com/riscv-software-src/riscv-perf-model/blob/55661600abe6604d3fc369da398e7677efdb1b1c/core/Rename.cpp#L292-L297

core/Rename.cpp
```diff
@@ -42,15 +42,23 @@ namespace olympia
         auto setup_map = [this] (core_types::RegFile reg_file, const uint32_t num_renames) {
                         uint32_t num_regs = 32;  // default risc-v ARF count
```
knute-sifive, line 43: Make this `const`

core/Rename.cpp
```diff
@@ -122,20 +130,26 @@ namespace olympia
...
-            auto const & original_dest = inst_ptr->getRenameData().getOriginalDestination();
-            --reference_counter_[original_dest.rf][original_dest.val];
-            // free previous PRF mapping if no references from srcs, there should be a new dest mapping for the ARF -> PRF
-            // so we know it's free to be pushed to freelist if it has no other src references
-            if(reference_counter_[original_dest.rf][original_dest.val] <= 0){
-                freelist_[original_dest.rf].push(original_dest.val);
+            const auto dest = dests[0];
+            const auto rf  = olympia::coreutils::determineRegisterFile(dest);
+            const auto num = dest.field_value;
+            if (num != 0 || rf != core_types::RF_INTEGER)
```
knute-sifive, line 136: I think this would be easier to read if this:
```suggestion
            if (!is_x0) 
```

core/Rename.cpp
```diff
@@ -271,18 +285,28 @@ namespace olympia
                 const auto & dests = renaming_inst->getDestOpInfoList();
                 for(const auto & src : srcs)
                 {
+                    const auto rf  = olympia::coreutils::determineRegisterFile(src);
+                    const auto num = src.field_value;
+                    // we check if src is RF_INTEGER x0, if so, we skip rename
+                    const bool is_x0 = (num == 0 && rf == core_types::RF_INTEGER);
+                    if (is_x0) {
+                        // if x0 is a data operand for LSU op, we need to set it in DataReg when we check in LSU
+                        // so we can still check the scoreboard, which will always return back ready for x0
```
knute-sifive, line 294: I don't think this is necessary.  The SB should never be checked for x0 -- in other words, you don't need to set that bit as a source for this instruction in the bit mask.

AaronGChan:
> This change seems a bit convoluted, would it not be better to refactor the allocation and freeing into separate functions, and act differently when the input register is X0.
> 
> Also, I can't find out where the scoreboard will always ready x0 reads?
> 
> https://github.com/riscv-software-src/riscv-perf-model/blob/55661600abe6604d3fc369da398e7677efdb1b1c/core/Rename.cpp#L292-L297

The scoreboard is marked here: https://github.com/riscv-software-src/riscv-perf-model/blob/55661600abe6604d3fc369da398e7677efdb1b1c/core/Rename.cpp#L107

And because we bypass any destinations of x0, we can assume the scoreboard is always marked ready. An alternative I'm considering is just storing whether a data operand in a LSU operation is x0, if so we just pass it through, without checking the scoreboard. 

With regards to refactoring, my initial thoughts were to keep as much of the original code as possible, but am considering ways to refactor it currently. Allocate and freeing are currently separated with regards to `renameInstructions_()` and `getAckFromROB_()`, but are you saying to have a more generic functions that we pass the operands through?

core/Inst.hpp
```diff
@@ -40,6 +40,7 @@ namespace olympia
                 uint32_t val = 0;
                 core_types::RegFile rf = core_types::RegFile::RF_INVALID;
                 mavis::InstMetaData::OperandFieldID field_id = mavis::InstMetaData::OperandFieldID::NONE;
+                bool x0 = false;
```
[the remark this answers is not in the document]
AaronGChan:
```
struct Reg {
    const uint32_t val;
    const core_types::RegFile rf;
    const mavis::InstMetaData::OperandFieldID field_id;
    const bool is_x0;

    Reg(uint32_t v, core_types::RegFile rf, mavis::InstMetaData::OperandFieldID field_id, bool is_x0)
        : val(v), rf(rf), field_id(field_id), is_x0(is_x0) {}
};
```

So I attempted to try and make the struct data members constant, but we then run into issues with the getter/setter functions because the copy constructor for the struct is deleted due to `const` on the data members.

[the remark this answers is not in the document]
klingaard: Ah, right.  I forgot you're replacing the internal members with new values.  Yeah, adding `const` will make the `Reg` struct ill-formed for assignment or move assignment.  Oh well.

core/Rename.cpp
AaronGChan: Resolved in [650f6c6](https://github.com/riscv-software-src/riscv-perf-model/pull/120/commits/650f6c6b24edbaac804d3a6219ddc783a829b8cc), added logic to just check `is_x0` and avoid even checking scoreboard bits.

core/Rename.cpp
AaronGChan: Addressed in [650f6c6](https://github.com/riscv-software-src/riscv-perf-model/pull/120/commits/650f6c6b24edbaac804d3a6219ddc783a829b8cc)

core/Rename.cpp
```diff
@@ -122,20 +134,27 @@ namespace olympia
...
-            auto const & original_dest = inst_ptr->getRenameData().getOriginalDestination();
-            --reference_counter_[original_dest.rf][original_dest.val];
-            // free previous PRF mapping if no references from srcs, there should be a new dest mapping for the ARF -> PRF
-            // so we know it's free to be pushed to freelist if it has no other src references
-            if(reference_counter_[original_dest.rf][original_dest.val] <= 0){
-                freelist_[original_dest.rf].push(original_dest.val);
+            const auto dest = dests[0];
+            const auto rf  = olympia::coreutils::determineRegisterFile(dest);
+            const auto num = dest.field_value;
+            const bool is_x0 = num != 0 || rf != core_types::RF_INTEGER;
```
klingaard, line 140: My brain hurts reading this...  do you mean `is_not_x0`?

Reversed logic:
```suggestion
            const bool is_x0 = num == 0 && rf != core_types::RF_INTEGER;
            if (false == is_x0)
```

Issue #1, opened 2022-09-23
Add a branch predictor


klingaard:
Need a branch predictor to Fetch. Open to considerations on the type. Should be parameterized, but a common API would be beneficial to keep the model generic.

Resolutions of the branches will require feedback from Retire and/or Execute . Typicality this can be done with ports.

arupc:
There is an example branch prediction uarchitecture described here:  https://docs.boom-core.org/en/latest/sections/branch-prediction/index.html . This can serve as an example to implement in Olympia

venky528:
Unable to get a view of current status of this ticket. @arupc could you please brief about its current status.

light2802:
We currently have made a timing sheet for the branch predictor based on the BOOM architecture. The plan is to model and implement the branch predictor APIs based on this timing sheet and integrate it with Olympia.

[[1] Branch Predictor timing sheet](https://docs.google.com/spreadsheets/d/1NNue_sBiCrq0VSX1MiECl7uqAFufwIzgk8j7p1gVFro/edit?usp=sharing)

egieske:
What is the most efficient way to determine if a Fetched Inst is a branch? Should we add a member to Inst like Inst::isStoreInst() or isLoadStoreInst()?
Similarly, how should a branch predictor determine branch instruction type, e.g., call, return, unconditional, etc?
Also, how should a branch predictor monitor the execution status of branch instructions? Inst::getStatus()==COMPLETE seems possible, but then the predictor will have to scan a list of in-flight branch Inst's every cycle to determine when it can update predictor internal state which may be inefficient.

klingaard:
> What is the most efficient way to determine if a Fetched Inst is a branch?
I have an answer to that question here: https://github.com/riscv-software-src/riscv-perf-model/issues/104#issuecomment-1785229711

> Similarly, how should a branch predictor determine branch instruction type, e.g., call, return, unconditional, etc?

Mavis supports multiple instruction type identifiers.  You can see them here: https://github.com/sparcians/mavis/blob/4f3fef891f9ddc5c371c27500d02596f21ea6fc8/mavis/InstMetaData.h#L30

So, if you want to know if a branch is JALR, you can add this to Inst.hpp:
```
     bool isJALR() const { return mavis_info_->isInstType(mavis::OpcodeInfo::InstructionTypes::JALR); }
```

> Also, how should a branch predictor monitor the execution status of branch instructions? 

[...]

PR #124, opened 2023-11-27
Initial implementation of branch prediction using a BTB

danbone:
Attempting to add some branch prediction logic to Olympia. I started with a branch target buffer.

Plan is to add a simple predictor, and a return address stack next. As this will require some form of recovery following speculative updates. Forming a foundation for further branch prediction modelling.

Early branch misprediction recovery could be done at a later date, but will require some more complex mechanism for restoring rename map before restarting.

Few points of interest:

- Adding flushing on misprediction has pushed the runtime up considerable, dhrystone takes 250s+ to simulate compared to 12secs on master. It's only doing 10k flushes, I'm surprised that it has had such an impact.
- The BTB model is basically just a clone of SimpleCache2 but the address decoder was a protected const, and I needed to change it. Any suggestion on a better way to do this?
- I intend to send all the flushed instructions to rename serially so to free up the renaming registers. This has an added bonus as it can be used to cancel the scoreboard callbacks. Unless there's a better suggestion?
- Should I update the JSON instruction generator? Will require new taken/target fields
- I want the branch prediction stuff encapsulated a bit more, I was thinking of having a new flush field in the instruction set on prediction, that can trigger flushes within the different modules. i.e. BTB miss could set flush to DECODE_FLUSH, branch misprediction sets EXECUTE_FLUSH. Would have to consider whether the instruction that triggers the flush is included or excluded from the flush. Thoughts?
- Any ideas on what tests to add?

### DIFF
--- a/core/BTB.hpp
+++ b/core/BTB.hpp
@@ -1,0 +1,277 @@
+#pragma once
+
+#include "sparta/simulation/Unit.hpp"
+#include "sparta/simulation/TreeNode.hpp"
+#include "sparta/simulation/ParameterSet.hpp"
+
+#include "sparta/utils/MathUtils.hpp"
+
+#include "cache/TreePLRUReplacement.hpp"
+#include "cache/BasicCacheItem.hpp"
+#include "cache/Cache.hpp"
+#include "cache/BasicCacheSet.hpp"
+
+#include "MiscUtils.hpp"
+
+
+namespace olympia
+{
+
+    /**
+     * @file   BTB.hpp
+     * @brief The Branch Target Buffer -- Cache like structure for branch instructions
+     *
+     * Based on SimpleCache, this unit defines a way associate cache like structure
+     * designed to be used as a way of steering fetch ahead.
+     *
+     * Each set covers an entire block of memory, there may be multiple branches with
+     * the same block address. We use a full tag to hit on the right line.
+     *
+     * We store the branch PC, target, type and a 2 bit saturating counter to predict
+     * direction on conditional branches.
+     */
+
+    class BTBEntry : public sparta::cache::BasicCacheItem
+    {
+    public:
+
+        enum class BranchType : std::uint16_t {
+            CONDITIONAL,
+            __FIRST = CONDITIONAL,
+            DIRECT,
+            INDIRECT,
+            RETURN,
+            __LAST
+        };
+
+        BTBEntry() :
+            valid_(false)
+        { };
+
+        BTBEntry(const BTBEntry &rhs) :
+            BasicCacheItem(rhs),
+            valid_(rhs.valid_),
+            target_(rhs.target_),
+            is_call_(rhs.is_call_),
+            branchtype_(rhs.branchtype_),
+            lhist_counter_(rhs.lhist_counter_)
+        { };
+
+        BTBEntry &operator=(const BTBEntry &rhs)
+        {
+            if (&rhs != this) {
+                BasicCacheItem::operator=(rhs);
+                valid_ = rhs.valid_;
+                target_ = rhs.target_;
+                is_call_ = rhs.is_call_;
+                branchtype_ = rhs.branchtype_;
+                lhist_counter_ = rhs.lhist_counter_;
+            }
+            return *this;
+        }
+
+        void reset(uint64_t addr)
+        {
+            setValid(true);
+            setAddr(addr);
+            resetLHistCounter_();
+        }
+
+        void setValid(bool v) { valid_ = v; }
+        bool isValid() const { return valid_; }
+
+        void setTarget(uint64_t t) { target_ = t; }
+        uint64_t getTarget() const { return target_; }
+
+        void setBranchType(BranchType branchtype, bool is_call = false)
+        {
+            sparta_assert(!is_call ||
+                          miscutils::isOneOf(branchtype, BranchType::INDIRECT,
+                                                         BranchType::DIRECT,
+                                                         BranchType::RETURN),
+                          "Invalid branch and call combination");
+            branchtype_ = branchtype;
+            is_call_ = is_call;
+        }
+
+        BranchType getBranchType() const { return branchtype_; }
+        bool isCall() const { return is_call_; }
+
+        bool predictDirection() {
+            if (branchtype_ == BranchType::CONDITIONAL)
+            {
+                return lhist_counter_ >= 0;
+            }
+            return true;
+        }
+
+        void updateDirection(bool taken)
+        {
+            if (taken)
+            {
+                lhist_counter_ = std::min(lhist_count_max_, ++lhist_counter_);
+            }
+            else
+            {
+                lhist_counter_ = std::max(lhist_count_min_, --lhist_counter_);
+            }
+        }
+
+
+    private:
+        bool                 valid_ = false;
+        uint64_t             target_;
+        bool                 is_call_ = false;
+        BranchType           branchtype_;
+        int32_t              lhist_counter_ = 0;
+        static constexpr int32_t lhist_count_max_ = 1; // 2 bit saturating counter.
+        static constexpr int32_t lhist_count_min_ = -2;
+
+        void resetLHistCounter_() { lhist_counter_ = 0; }
+
+    };
+
+
+    class BTBAddrDecoder : public sparta::cache::AddrDecoderIF
+    {
+    public:
+        BTBAddrDecoder(uint32_t entries, uint32_t stride, uint32_t num_ways)
+        {
+            index_mask_ = (entries/num_ways)-1;
+            index_shift_ = sparta::utils::floor_log2(stride);
+        }
+        virtual uint64_t calcTag(uint64_t addr) const { return addr; }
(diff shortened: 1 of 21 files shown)


core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+            if (&rhs != this) {
+                BasicCacheItem::operator=(rhs);
+                valid_ = rhs.valid_;
+                target_ = rhs.target_;
+                is_call_ = rhs.is_call_;
+                branchtype_ = rhs.branchtype_;
+                lhist_counter_ = rhs.lhist_counter_;
+            }
+            return *this;
+        }
```
klingaard, line 71: You can do this:
```
```suggestion
        BTBEntry() = default;
        BTBEntry(const BTBEntry &rhs) = default;
        BTBEntry &operator=(const BTBEntry &rhs) = default;
```

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+
+        void reset(uint64_t addr)
+        {
+            setValid(true);
+            setAddr(addr);
+            resetLHistCounter_();
+        }
```
klingaard, line 78: I've never been a fan of `reset` methods.  Instead, you can do this after adding a new constructor that accepts an address:
```
   btb_entries_[idx] = BTBEntry(addr);
```

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+    public:
+        BTBAddrDecoder(uint32_t entries, uint32_t stride, uint32_t num_ways)
+        {
+            index_mask_ = (entries/num_ways)-1;
+            index_shift_ = sparta::utils::floor_log2(stride);
+        }
+        virtual uint64_t calcTag(uint64_t addr) const { return addr; }
+        virtual uint32_t calcIdx(uint64_t addr) const { return (addr >> index_shift_) & index_mask_; }
+        virtual uint64_t calcBlockAddr(uint64_t addr) const { return addr; }
+        virtual uint64_t calcBlockOffset(uint64_t addr) const { return 0;}
```
klingaard, line 146: Are these new `virtual` methods or overrides from `AddrDecoderIF`?  If overrides, remove the `virtual` keyword and add `override` to the end.

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+    private:
+        uint32_t index_mask_;
+        uint32_t index_shift_;
```
klingaard, line 149: Make these `const`

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+    };
+
+
+    class BTBAddrDecoder : public sparta::cache::AddrDecoderIF
+    {
+    public:
+        BTBAddrDecoder(uint32_t entries, uint32_t stride, uint32_t num_ways)
+        {
+            index_mask_ = (entries/num_ways)-1;
+            index_shift_ = sparta::utils::floor_log2(stride);
```
klingaard, line 141: Move these to the initializer list

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+            return cache_->getCacheSet(addr).getItemForReplacementWithInvalidCheck();
+        }
+
+        /**
+         *\return Pointer to line with addr.  nullptr is returned if not fould
+        */
+        BTBEntry * getLine(uint64_t addr)
+        {
+            return cache_->getItem(addr);
+        }
```
klingaard, line 204: Shouldn't this be a `const` method returning a `const` entry?

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+
+        // Allocate 'line' as having the new 'addr'
+        // 'line' doesn't know anything about NT
+        void allocateWithMRUUpdate(BTBEntry &line,
+                                    uint64_t   addr)
+        {
+            line.reset( addr );
+            touchMRU( line );
+        }
```
klingaard, line 233: Why does this take a `line` and then reset it?  How about it creates one and returns it?

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+
+
+        void invalidateLineWithLRUUpdate(BTBEntry &line)
+        {
+            static const uint64_t addr=0;
+            line.reset( addr );
+            line.setValid( false );
+            touchLRU( line );
+        }
```
klingaard, line 242: Same question here as above.

core/BTB.hpp
```diff
@@ -0,0 +1,277 @@
...
+                }
+                set_it->getReplacementIF()->reset();
+
+            }
+        }
+
+        /**
+         * determine if there are any open ways in the set.
+         */
+        bool hasOpenWay(const uint64_t addr)
```
klingaard, line 260: Can this method be made `const`?

core/Fetch.cpp
```diff
@@ -48,9 +50,49 @@ namespace olympia
...
                                                          skip_nonuser_mode_);
 
+        // Setup BTB
+        btb_ = getContainer()->getChild("btb")->getResourceAs<olympia::BTB>();
+
         fetch_inst_event_->schedule(1);
     }
 
+
+    bool Fetch::predictInstruction_(InstPtr inst)
```
klingaard, line 60: `const InstPtr & inst` to avoid the copy

core/Fetch.cpp
```diff
@@ -48,9 +50,49 @@ namespace olympia
...
+    {
+        if (btb_->isHit(inst->getPC()))
+        {
+            btb_hits_++;
```
klingaard, line 64: Use pre-increment to avoid the copy (although the compiler will do that for you).

core/BTB.hpp
klingaard, line 1: For the API, I suggest you don't use references as the parameters, but instead use pointers:
```
(const BTBEntry * line)
```
Reason for this, look at this example of buggy code:
```
    auto btb_entry = btb_->getEntry();
    doSomeStuffToUpdateEntry_(&btb_entry);
```
Does `getEntry` return reference or a copy?  The `auto` might make a copy and the call to `doSomeStuffToUpdateEntry_` will not update the original.

core/Rename.cpp
klingaard, line 1: So the changes here reset the entirety of rename, preventing support for partial flushes.  :thinking: -- would you be up to updating the flushing criteria to include those instructions being flushed?

klingaard: > Early branch misprediction recovery could be done at a later date, but will require some more complex mechanism for restoring rename map before restarting.

Ah, you already mentioned this...

core/BTB.hpp
danbone: I'm not sure I can do this (unless I'm not understanding something, fairly new to C++).

The problem is that we're using the [Cache ](https://github.com/sparcians/map/blob/master/sparta/cache/Cache.hpp)and [BasicCacheSet](https://github.com/sparcians/map/blob/master/sparta/cache/BasicCacheSet.hpp) as the underlying data structures. The BTB acts as a shim.

The Cache returns pointers to the BTBEntry objects stored within a vector inside BasicCacheSet, the user can then manipulate the object using the pointer, but there isn't a way to put a new object into the vector down in the Set class.

Is there a way I can use the constructor you suggested to copy into the pointer object on the RHS?

(I tried to experiment a bit, but couldn't get the syntax right - complains that the btb_entry is a pointer and if use new then I think you'd get a memory leak?)

arupc:
Thank you for initiating this. 

To provide more history, per previous discussion, we intend to add branch prediction support via an API, along with an implementation of simple branch predictor. This would enable use of a simple predictor while allowing for implementation of other predictors. There's also an interest in re-using existing branch predictor implementations

Please see  https://github.com/riscv-software-src/riscv-perf-model/issues/1 

I will start a discussion item on this topic. Let's collaborate via the discussion thread and review a proposal for branch prediction API

klingaard:
From our discussion in the SIG yesterday:

* This is the first PR that actually hammers on the flushing mechanism (and actually found a bug). 
* While this doesn't use the proposed branch API from @arupc and team, it's not a show-stopper for moving forward with this PR
* The flushing/redirect occurs in Decode, but only for a BTB miss (never seen before branch).  When this occurs the _entire pipe_ is flushed.  Not sure how safe this is right.  Might behoove us to actually implement a proper redirect from execution instead.
* The PR is missing some fundamental testing.  Specifically, we'd like to see a unit test _just for the BTB_ as it stands so it can be grown over time.  Also, augmentation of the JSON format might be handy to support taken/not taken branch behavior to test flushing, etc.

danbone:
- Updated branch with a new FlushingCriteria definition for determining which instructions to flush (inclusive/exclusive flush)
- Remove code which set scoreboard bits on flush which then led to fix flushing instructions from the issue queues in the executepipe/lsu that have outstanding scoreboard callbacks
- Updated rename map recovery to use iterate over the flushed instructions from the ROB and fix up the map/free registers.
- Moved ROB allocation from dispatch to rename. Ran into a race condition where instructions updated the rename map, but weren't included in the flush because they were stalled in dispatch and therefore hadn't been appended to the reorder buffer.

danbone:
Moving forward following the discussion in our last meeting on the 20th of December. I think we should separate out the changes needed for flushing. Then rebase the branch prediction API branch on top of it.

PR #137, opened 2024-01-04
Add execute flushing

danbone:
- Fix pipeline flush mechanisms by removing inflight instructions inside functional units, recovering the rename map, and adding trace rewind functionality
- Add FlushCriteria class for requesting flushes to and from the FlushManager
- Update FlushManager to handle and arbitrate between multiple flush sources
- Add flush request port from ExecutePipe to FlushManager intended to be used for mispredicted branches
- Delay completing instructions in ExecutePipe by a cycle to fix execute flush race conditions
- Add branch identification methods for Inst
- Add test to send random flushes from the branch unit

### DIFF
--- a/core/ExecutePipe.cpp
+++ b/core/ExecutePipe.cpp
@@ -234,7 +234,7 @@ namespace olympia
             auto inst_ptr = *delete_iter;
 
             // Remove flushed instruction from issue queue and clear scoreboard callbacks
-            if (criteria.flush(inst_ptr))
+            if (criteria.includedInFlush(inst_ptr))
             {
                 issue_queue_.erase(delete_iter);
 
@@ -258,7 +258,7 @@ namespace olympia
         {
             auto delete_iter = ready_queue_iter++;
             auto inst_ptr = *delete_iter;
-            if (criteria.flush(inst_ptr))
+            if (criteria.includedInFlush(inst_ptr))
             {
                 ready_queue_.erase(delete_iter);
                 ILOG("Flush Instruction ID: " << inst_ptr->getUniqueID() << " from ready queue");
@@ -268,7 +268,7 @@ namespace olympia
         // Cancel outstanding instructions awaiting completion and
         // instructions on their way to issue
         auto flush = [criteria](const InstPtr & inst) -> bool {
-            return criteria.flush(inst);
+            return criteria.includedInFlush(inst);
         };
         issue_inst_.cancel();
         complete_inst_.cancelIf(flush);
--- a/core/ExecutePipe.cpp
+++ b/core/ExecutePipe.cpp
@@ -43,7 +43,7 @@ namespace olympia
         // Complete should come before issue because it schedules issue with a 0 cycle delay
         // issue should always schedule complete with a non-zero delay (which corresponds to the
         // insturction latency)
-        complete_inst_ >> issue_inst_;
+        execute_inst_ >> issue_inst_;
 
         ILOG("ExecutePipe construct: #" << node->getGroupIdx());
 
@@ -138,7 +138,7 @@ namespace olympia
 
         ++total_insts_issued_;
         // Mark the instruction complete later...
-        complete_inst_.preparePayload(ex_inst_ptr)->schedule(exe_time);
+        execute_inst_.preparePayload(ex_inst_ptr)->schedule(exe_time);
         // Mark the alu as busy
         unit_busy_ = true;
         // Pop the insturction from the ready queue and send a credit back to dispatch
@@ -148,10 +148,9 @@ namespace olympia
     }
 
     // Called by the scheduler, scheduled by complete_inst_.
-    void ExecutePipe::completeInst_(const InstPtr & ex_inst)
+    void ExecutePipe::executeInst_(const InstPtr & ex_inst)
     {
-        ex_inst->setStatus(Inst::Status::COMPLETED);
-        ILOG("Completing inst: " << ex_inst);
+        ILOG("Executed inst: " << ex_inst);
 
         // set scoreboard
         if(SPARTA_EXPECT_FALSE(ex_inst->isTransfer()))
@@ -189,6 +188,17 @@ namespace olympia
         if (ready_queue_.size() > 0) {
             issue_inst_.schedule(sparta::Clock::Cycle(0));
         }
+
+        // Schedule completion
+        complete_inst_.preparePayload(ex_inst)->schedule(1);
+
+    }
+
+    // Called by the scheduler, scheduled by complete_inst_.
+    void ExecutePipe::completeInst_(const InstPtr & ex_inst)
+    {
+        ex_inst->setStatus(Inst::Status::COMPLETED);
+        ILOG("Completing inst: " << ex_inst);
     }
 
     void ExecutePipe::flushInst_(const FlushManager::FlushingCriteria & criteria)
@@ -198,10 +208,10 @@ namespace olympia
         // Flush instructions in the issue queue
         uint32_t credits_to_send = 0;
 
-        auto iter = issue_queue_.begin();
-        while (iter != issue_queue_.end())
+        auto issue_queue_iter = issue_queue_.begin();
+        while (issue_queue_iter != issue_queue_.end())
         {
-            auto delete_iter = iter++;
+            auto delete_iter = issue_queue_iter++;
 
             auto inst_ptr = *delete_iter;
 
@@ -214,7 +224,7 @@ namespace olympia
 
                 ++credits_to_send;
 
-                ILOG("Flush Instruction ID: " << inst_ptr->getUniqueID());
+                ILOG("Flush Instruction ID: " << inst_ptr->getUniqueID() << " from issue queue");
             }
         }
 
@@ -225,27 +235,34 @@ namespace olympia
         }
 
         // Flush instructions in the ready queue
-        auto flush = [criteria](const InstPtr & inst) -> bool {
-            return criteria.flush(inst);
-        };
-        ready_queue_.erase(std::remove_if(ready_queue_.begin(),
-                                          ready_queue_.end(),
-                                          flush),
-                           ready_queue_.end());
+        auto ready_queue_iter = ready_queue_.begin();
+        while (ready_queue_iter != ready_queue_.end())
+        {
+            auto delete_iter = ready_queue_iter++;
+            auto inst_ptr = *delete_iter;
+            if (criteria.flush(inst_ptr))
+            {
+                ready_queue_.erase(delete_iter);
+                ILOG("Flush Instruction ID: " << inst_ptr->getUniqueID() << " from ready queue");
+            }
+        }
 
         // Cancel outstanding instructions awaiting completion and
         // instructions on their way to issue
+        auto flush = [criteria](const InstPtr & inst) -> bool {
+            return criteria.flush(inst);
+        };
+        issue_inst_.cancel();
         complete_inst_.cancelIf(flush);
-        if(complete_inst_.getNumOutstandingEvents() == 0) {
+        execute_inst_.cancelIf(flush);
+        if(execute_inst_.getNumOutstandingEvents() == 0) {
             unit_busy_ = false;
             collected_inst_.closeRecord();
+            if (!ready_queue_.empty()) {
+                // issue non-flushed instructions
+                issue_inst_.schedule(sparta::Clock::Cycle(1));
         }
-
-        if (ready_queue_.empty())
-        {
-            issue_inst_.cancel();
         }
-
     }
 
     ////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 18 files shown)


danbone:
@knute-sifive few questions for you

1. 
For the flush testing mode we spoke about in the call yesterday. Were you thinking to use a parameter in the ExecutePipe?

2. 
I've run into a race condition where a branch mispredicts, completes and retires in the same cycle, before the flush is seen. Meaning any younger instruction that is retired alongside should have actually been flushed

See this log, where PID: 1975 is correctly refetched following the flush but also retired just before it.
```{0000004137 00004137 top.cpu.core0.execute.br0 info} completeInst_: mispredicted branch uid: 2551  COMPLETED 113fe pid: 1974 'bne	x5,x7 +0xb2'  was actually not-taken
{0000004137 00004137 top.cpu.core0.rob info} retireInstructions_: retiring uid: 2551    RETIRED 113fe pid: 1974 'bne	x5,x7 +0xb2' 
{0000004137 00004137 top.cpu.core0.rob info} retireInstructions_: retiring uid: 2552    RETIRED 11402 pid: 1975 'bne	x12,x13 +0x3c' 
{0000004138 00004138 top.cpu.core0.rename info} getAckFromROB_: Retired instruction: uid: 2551    RETIRED 113fe pid: 1974 'bne	x5,x7 +0xb2' 
{0000004138 00004138 top.cpu.core0.rename info} getAckFromROB_: Retired instruction: uid: 2552    RETIRED 11402 pid: 1975 'bne	x12,x13 +0x3c' 
{0000004138 00004138 top.cpu.core0.rob info} handleFlush_: flushing uid: 2574 DISPATCHED 1144a pid: 1997 'slli	x14,x12, SHAMT=0x20' 
```

I think we need to fix it so that we don't retire instructions that completed that cycle, and there's a few ways of doing this:
a. Add a sharedData object inside Inst that delays the completed status.
b. Add a completed event inside Inst and have retirement check if the event is scheduled before retiring (event should only be scheduled on the cycle it's completing)
c. Delay setting status to complete inside ExecutePipe and LSU

Any input?

danbone:
As the race condition was between the flush coming from the execute and the instruction changing to completed status. I decided to separate changing the status to completed and the execution of the instruction.

klingaard:
Thanks for doing this!

> For the flush testing mode we spoke about in the call yesterday. Were you thinking to use a parameter in the ExecutePipe?

This is related to the "random" flushing idea (for testing), right?  Here are my thoughts:
1. Since the flushes are instigated mostly (if not always) from the ROB, add the parameter there.  It's simple integer-based parameter, `enable_random_flushes` that takes a random key.  If non-zero, it's enabled with that key.  
2. Add a few test with some random keys (maybe using the `$RAMDOM` variable from bash)

Glad you got the race condition rectified.

core/Inst.hpp
```diff
@@ -148,6 +163,13 @@ namespace olympia
         void setLast() { is_last_ = true; }
         bool getLast() const { return is_last_; }
 
+        // Set the STF iterator to rewind trace when flushing
+        void setSTFIterator(stf::STFInstReader::iterator it) { trace_it_ = it; }
+        stf::STFInstReader::iterator getSTFIterator() const { return std::get<stf::STFInstReader::iterator>(trace_it_); }
+
+        void setJSONIterator(uint64_t it) { trace_it_ = it; }
+        uint64_t getJSONIterator() const { return std::get<uint64_t>(trace_it_); }
```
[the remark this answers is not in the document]
danbone: Can you give me a bit more guidance here? 

Where should I create the class to? Inside Inst?

If it was execution driven we would just use the PC to rewind?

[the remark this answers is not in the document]
klingaard: Sure. first thought: a local member variable that is a `tuple`.  Add a template method that returns the different type.

(Dunno if this will do the trick -- just brain storming)

```
// Defined in Generator header files
enum Generator { STF, JSON };
using JSONIterator = uint64_t;

// Defined in Inst.hpp
public:
    template<Generator gen>
    auto getInstIterator() const {
        return std::get<gen>(rewind_iterator_);
    }

private:
    std::tuple<stf::STFInstReader::iterator, JSONIterator> rewind_iterator_;
```
Usage:
```
   inst_ptr->getInstIterator<Generator::JSON>();
```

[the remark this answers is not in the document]
klingaard: > If it was execution driven we would just use the PC to rewind?

Not that simple.  Depending on how you implement exe mode, it might need more information than which PC do I need to go back to.  For example, if the functional components keep track of uncommitted changes (its own renames), then the functional engine might need more than the PC.  This is especially true if we're deep into a loop and the functional model has seen that PC more than once...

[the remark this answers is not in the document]
danbone: OK makes sense.

Is there a reason why a tuple is preferred over a union like std::variant

danbone:
I hope I've address your comments above.

I played around with the ROB flushing but I found that doing it in the branch unit found more bugs. This should be a temporary thing to plug the gap until branch prediction is implemented.

For the ROB flushing, I could still put it in. What I did based on your comments was to use a hidden parameter that acted like a UID to trigger a flush from. Obviously this would only cause 1 flush per run, so it didn't have a lot of coverage compared to the branch unit flushing.

core/Inst.hpp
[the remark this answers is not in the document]
knute-mips: Nope, no reason.  That's actually better since we'll only support 1 input at a time...

danbone:
Just noticed I forgot to change the FlushManager::FlushCriteria.flush method as requested

core/Rename.cpp
```diff
@@ -184,6 +195,72 @@ namespace olympia
...
+
+                        // free renamed PRF mapping when reference counter reaches zero
+                        auto const & renamed_dest = inst_ptr->getRenameData().getDestination();
+                        --reference_counter_[renamed_dest.rf][renamed_dest.val];
+                        if(reference_counter_[renamed_dest.rf][renamed_dest.val] <= 0){
+                            freelist_[renamed_dest.rf].push(renamed_dest.val);
+                        }
+                    }
+                    else
+                    {
```
AaronGChan, line 228: We probably can remove this empty else right? If an instruction destination is x0, we don't need to rewind anything in terms of flushing.

core/ExecutePipe.cpp
```diff
@@ -170,6 +185,17 @@ namespace olympia
             scoreboard_views_[reg_file_]->setReady(dest_bits);
         }
 
+        // Testing mode to inject random branch misprediction to stress flushing mechanism
+        if (enable_random_misprediction_)
+        {
+            if (ex_inst->isBranch() && (std::rand() % 20) == 0)
```
knute-mips, line 191: This is awesome, but can a failure be reproduced?  Probably should add a parameter to set a seed...

Also, can the 20 be parameterized as well?  To increase the chances (for interesting experiments).

core/FlushManager.hpp
```diff
@@ -38,7 +42,66 @@ namespace olympia
...
+
+            bool isInclusiveFlush() const
+            {
+                static const std::map<FlushCause, bool> inclusive_flush_map = {
+                    {FlushCause::TRAP,                 true},
+                    {FlushCause::MISFETCH,             true},
+                    {FlushCause::MISPREDICTION,        false},
+                    {FlushCause::TARGET_MISPREDICTION, false},
+                    {FlushCause::POST_SYNC,            false}
+                };
```
knute-mips, line 79: This is fine for now... but I wonder if this should be the decision of the instruction causing the flush...

core/Inst.cpp
```diff
@@ -23,7 +44,10 @@ namespace olympia
         is_transfer_(miscutils::isOneOf(inst_arch_info_->getTargetPipe(),
                                         InstArchInfo::TargetPipe::I2F,
                                         InstArchInfo::TargetPipe::F2I)),
-
+        is_branch_(opcode_info_->isInstType(mavis::OpcodeInfo::InstructionTypes::BRANCH)),
+        is_condbranch_(opcode_info_->isInstType(mavis::OpcodeInfo::InstructionTypes::CONDITIONAL)),
+        is_call_(isCallInstruction(opcode_info)),
+        is_return_(isReturnInstruction(opcode_info)),
```
knute-mips, line 50: NICE!

knute-mips (approved):
LGTM -- Aaron has a comment worth addressing.

core/Rename.cpp
danbone: Good catch, I must have missed it in the merge - removed.

core/FlushManager.hpp
danbone: Did think about doing it that way, but then I thought against it as I figured it would be more robust tying the cause and inclusiveness of the flush together. I don't feel too strongly about it, but perhaps we could keep it this way for now?

knute-mips: Yeah, I actually like this approach, so feel free to keep it this way for now.

klingaard (approved):

klingaard:
Will merge this tomorrow morning CST.

core/ExecutePipe.cpp
kathlenehurt-sifive: You could change the parameter `enable_random_misprediction` to `random_misprediction_rate` and default it to zero.

core/ExecutePipe.cpp
```diff
@@ -180,43 +206,108 @@ namespace olympia
...
+
+                scoreboard_views_[reg_file_]->clearCallbacks(inst_ptr->getUniqueID());
+
                 ++credits_to_send;
-            }
-            else {
-                ++it;
+
+                ILOG("Flush Instruction ID: " << inst_ptr->getUniqueID() << " from issue queue");
             }
```
kathlenehurt-sifive, line 246: Is the issue queue age ordered? If so, can't we just break when we find the first instruction that is not included in the flush?

core/ExecutePipe.cpp
```diff
@@ -180,43 +206,108 @@ namespace olympia
...
+            }
+        }
+    }
+
+    ////////////////////////////////////////////////////////////////////////////////
+    // Regular Function/Subroutine Call
+    ////////////////////////////////////////////////////////////////////////////////
+
+    // Append instruction into issue queue
+    void ExecutePipe::appendIssueQueue_(const InstPtr & inst_ptr)
```
kathlenehurt-sifive, line 291: The [Sparta Buffer class](https://github.com/sparcians/map/blob/8b16b216de9960cacdc22d78820165ab86d6de58/sparta/sparta/resources/Buffer.hpp) might be good to use for the issue queue. It allows for append and erase from anywhere in the buffer. It has its own implementation of the two methods you have here (`appendIssueQueue_` and `popIssueQueue_`) so you wouldn't need to provide your own implementation.

core/Fetch.cpp
```diff
@@ -105,16 +106,28 @@ namespace olympia
...
-        ILOG("Fetch: receive flush on new_addr=0x"
-             << std::hex << new_addr << std::dec);
+    // Called from FlushManager via in_fetch_flush_redirect_port
+    void Fetch::flushFetch_(const FlushManager::FlushingCriteria &criteria)
+    {
+        ILOG("Fetch: received flush " << criteria);
+
+        auto flush_inst = criteria.getInstPtr();
+
+        // Rewind the tracefile
```
kathlenehurt-sifive, line 116: You can remove the if-statement here by passing `!criteria.isInclusiveFlush()` as the second parameter to `reset` instead.

kathlenehurt-sifive:
> Will merge this tomorrow morning CST.

None of my comments need to be addressed before merging. They can be resolved in another PR.

core/ExecutePipe.cpp
danbone: It should be, I avoided inferring the ordering though. I figured it's worth the extra cycles to make flushing more flexible.  i.e. if we wanted to implement multi-threading.

core/ExecutePipe.cpp
danbone: I forgot to reply back to this last night.

With std::rand, if the seed isn't set then it's always seeded with 1 according to google so should be reproducible (for now)

I think the random_misprediction_rate is a good idea. Something like tuning mispredictions per 1k branches.
I don't have a good knowledge of the C++ random libraries, but I saw I could use the discrete_distribution class. Happy to do this in another PR, as I think it'll need some back and forth discussion.

klingaard:
Daniel, would you like more time to address Kathlene's comments or would you prefer to do that in another PR?

danbone:
Another PR please

core/ExecutePipe.cpp
danbone: Might be something @AaronGChan can pick up in the issue queue changes

core/ExecutePipe.cpp
kathlenehurt-sifive: Ah! I didn't think of that. Makes sense.

PR #138, opened 2024-01-08
Fix end of simulation empty buffer checks in LSU

danbone:
I noticed that the buffer empty termination checks weren't working as intended.


danbone:
Thanks, I don't have write access will you merge it? (not sure what the process is)

klingaard:
Still new on the process.  Currently those folks that are directly on this project can merge.  Will need to do some chin-scratching with Arup to determine the "right of passage" for directly contributing.   This is new to me...

PR #139, opened 2024-01-17
Branch Prediction API

arupc-vmicro:
A code introducing branch prediction API. Still a draft PR as I am working through a compilation issue. 

Please feel free to review and provide feedback.


klingaard:
@bdutro you might have some input...

danbone:
Hi Arup, following up on our discussion last night. Could you help me understand how the API will provide configurability, and support speculative state updates?

For configurability, how do we pass in runtime parameters. Furthermore how can we enable reuse of predictor components.
A simple example could be that we have two different branch predictors, one is a bimodal table (similar to what you have implemented), the other is a global history predictor such as gshare (i.e. a branch history table that indexed through a hash of the global history and the PC).  Both predictors have different parameters, and you would likely want to swap them in and out but use the same BTB (and RAS/indirect predictor).

Like the Cache model allows users to change the address decoder, do we need similar methods for changing the components within the branch predictor i.e. the conditional branch predictor, the indirect target predictor, BTB and return address stack.

The other point is about supporting predictors which have internal speculative state. For example a global history predictor has the global history register which needs to be recovered following a flush. Similarly some predictors require meta data about the prediction to be available for the update, TAGE requires the provider table, and alternate prediction for example.

Finally we did speak about how to fit a block based BTB, like you've implemented, into a trace execution model. Branch predictors work in addresses, using trace is difficult because we have to pair instructions with their predictions based on their address. How do we handle malformed trace, or trace that has unexpected change of flow (i.e. interrupt/exception).

danbone:
Knute mentioned looking at other simulators. I've used gem5, but I'm also aware of champsim.

**gem5 o3 model**
The branch predictor is an instance within the fetch unit. It contains the BTB, RAS, conditional and indirect target predictors. However the BTB, indirect target predictor and RAS are all objects within the branch unit, which can be swapped out, or disabled entirely.

```
Fetch
    ├── BranchPredictor
    │   ├── btb
    │   ├── ras
    │   ├── indirectBranchPred
    │   └── IndirectPredictor
```

As for the API you can read the documentation [here](https://doxygen.gem5.org/release/current/classgem5_1_1branch__prediction_1_1BPredUnit.html#ae3543b5202a619a43c99664b5fc87bb6) for details but I'll summarize.

The interface with Fetch is just 3 methods: predict, squash(flush) and update. 
Predict is called for each instruction fetched, the instruction itself is passed through the API. The predictor relies on a sequence number (seqNum) to map predictions to squashes/updates, as the updates/history is stored within the class.

The BTB/RAS/Indirect predictor objects can be swapped out in the build configuration as long as they're a subclass of the original implementation. 

The conditional predictor on the other hand, must be a subclass of the branch predictor unit itself. Many leave the base predict, squash and update implementations, and instead override the methods referenced in those. Mainly the lookup*, updateHistories*, update (internal one), and squash.

*Both referenced by predict.

**ChampSim**
In ChampSim (looking at the ooo_cpu) it separates the BTB and BranchPredictor into two separate classes, though the BTB contains the RAS. The ooo_cpu does a prediction for each instruction fetched (as far as I can tell). It also validates the predictions against the trace, mispredictions stall the pipeline until their resolved.

Both the BTB and BranchPredictor take the PC to produce a prediction. The update methods take the PC, target, direction and branch type.


Going back to Olympia I think the makePrediction/doUpdate methods are fine, given how generic they are. But I do think more thought is needed on how to integrate these, providing parameterization and enable statistic collection.

It might be better to have separate objects for the conditional predictor, indirect target, BTB and RAS. With another API defined for each of them. And a simpleBranchPredictor implementation which orchestrates predictions, updates, flushes across those objects.

arupc-vmicro:
> For configurability, how do we pass in runtime parameters. Furthermore how can we enable reuse of predictor components.
A simple example could be that we have two different branch predictors, one is a bimodal table (similar to what you have implemented), the other is a global history predictor such as gshare (i.e. a branch history table that indexed through a hash of the global history and the PC). Both predictors have different parameters, and you would likely want to swap them in and out but use the same BTB (and RAS/indirect predictor).

Addressing this part of question/comment: We can do this using ResourceFactories, which are used to configure the tree nodes. As an example, please see files `core/CPUFactories.hpp`. While sparta units such fetch, decode use a default factory, the unit execute uses a custom factory defined in `core/Execute.cpp`, which goes through the core extension yaml and creates execution units.

Extending that idea, we can do the following:
* use a simple branch predictor or a more complex one (like specified in https://github.com/riscv-software-src/riscv-perf-model/issues/143) based on a command line parameter
* for a more complex branch predictor with sub-predictors, select one of multiple available indirect predictor, based on parameters

arupc-vmicro:
> Like the Cache model allows users to change the address decoder, do we need similar methods for changing the components within the branch predictor i.e. the conditional branch predictor, the indirect target predictor, BTB and return address stack.

IMO, the API should be generic enough to support simplest of the predictors, which may not have different sub-predictors as as long as API does not prevent or create undue burden on implementing more complicated predictors with sub-predictors.


> The other point is about supporting predictors which have internal speculative state. For example a global history predictor has the global history register which needs to be recovered following a flush. Similarly some predictors require meta data about the prediction to be available for the update, TAGE requires the provider table, and alternate prediction for example.

Can we not use the updatePredictor method with appropriate input (sub-classed from DefaultUpdate or defined separately) to restore/change predictor states, as needed?

> Finally we did speak about how to fit a block based BTB, like you've implemented, into a trace execution model. Branch predictors work in addresses, using trace is difficult because we have to pair instructions with their predictions based on their address. How do we handle malformed trace, or trace that has unexpected change of flow (i.e. interrupt/exception).

We do have to add proper mechanism to handle traces. I plan to address this later.

arupc-vmicro:
> I think the makePrediction/doUpdate methods are fine, given how generic they are. But I do think more thought is needed on how to integrate these, providing parameterization and enable statistic collection.

I agree and issue #143 will give us the opportunity to work on these aspects. 

> It might be better to have separate objects for the conditional predictor, indirect target, BTB and RAS. With another API defined for each of them, and then a BranchPredict implementation which orchestrates predictions, updates, flushes across those objects.

I am happy to build on top of the currently proposed API, once reviewed and finalized, to create sub-predictor-specific APIs, if that's needed.

My suggestion would be that we work towards implementing #143 using the proposed branch prediction API and Sparta subunits to create, parameterize sub-predictors and combine them. As we work through it, we can discuss how can we keep interfaces generic enough to be APIs

danbone:
On a different note do we need to discuss the directory structure going forward for when others implement their own predictors? Do we need a separate branchpred directory under core?

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+    // branch_idx can vary from 0 to (FETCH_WIDTH - 1)
+    uint32_t branch_idx;
+    // predicted target PC 
+    uint64_t predictedPC;
+};
+
+class DefaultUpdate
+{
+public:
+    uint64_t FetchPC;
```
rajatbhatia1, line 55: Please initialize all members

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+class BTBEntry 
+{
+public:
+    // use of BTBEntry in std:map operator [] requires default constructor
+    BTBEntry(uint32_t bidx=0, uint64_t predPC=0) :
+        branch_idx(bidx),
+        predictedPC(predPC)
+    {} 
+    uint32_t branch_idx;
```
rajatbhatia1, line 76: branch_idx {0};
predictedPC {0};

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+public:
+    // PC of first instruction of fetch packet
+    uint64_t FetchPC;
+};
+
+class BTBEntry 
+{
+public:
+    // use of BTBEntry in std:map operator [] requires default constructor
+    BTBEntry(uint32_t bidx=0, uint64_t predPC=0) :
```
rajatbhatia1, line 72: Replace with BTBEntry(uint32_t bidx, uint64_t predPC)

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+ * implementations of getPrediction and updatePredictor operations.
+ * */
+#pragma once
+
+#include <cstdint>
+#include <map>
+#include "sparta/utils/SpartaAssert.hpp"
+
+template <class PredictionT, class UpdateT, class InputT>
+class BranchPredictorIF
```
rajatbhatia1, line 33: Suggest this be in a separate header. User of BranchPredictorIF is not obligated to use and hence need not link to SimpleBranchPredictor

rajatbhatia1 (commented):
Please put everything in separate namespace

core/BranchPred.cpp
```diff
@@ -0,0 +1,70 @@
...
+}
+
+DefaultPrediction SimpleBranchPredictor::getPrediction(const DefaultInput & input) {
+    bool predictTaken;
+    if (branch_history_table_.find(input.FetchPC) != branch_history_table_.end()) {
+        predictTaken = (branch_history_table_[input.FetchPC] > 1);
+    } else {
+        predictTaken = false;
+        // add a new entry to BHT, biased towards not taken
+        branch_history_table_.insert(std::pair<uint64_t, uint8_t>(input.FetchPC, 1));
```
rajatbhatia1, line 46: Is there a reason this is initialized to 1 and not 0?
branch_history_table_.insert(std::pair<uint64_t, uint8_t>(input.FetchPC, 0));

core/BranchPred.cpp
```diff
@@ -0,0 +1,70 @@
...
+    } else {
+        predictTaken = false;
+        // add a new entry to BHT, biased towards not taken
+        branch_history_table_.insert(std::pair<uint64_t, uint8_t>(input.FetchPC, 1));       
+    }
+
+    DefaultPrediction prediction;
+    if (branch_target_buffer_.find(input.FetchPC) != branch_target_buffer_.end()) {
+        // BTB hit
+        BTBEntry btb_entry = branch_target_buffer_[input.FetchPC];
```
rajatbhatia1, line 52: Use reference to prevent copy of btb entry
BTBEntry &btb_entry = branch_target_buffer_[input.FetchPC];

core/BranchPred.cpp
```diff
@@ -0,0 +1,70 @@
...
+        prediction.branch_idx = btb_entry.branch_idx;
+        if (predictTaken) {
+            prediction.predictedPC = btb_entry.predictedPC;
+        } else {
+            // fall through address
+            prediction.predictedPC = input.FetchPC + prediction.branch_idx + BYTES_PER_INST;
```
rajatbhatia1, line 58: This limits it to fix width instruction sets only (in this case 4 bytes per instruction). Should be ok as long as we document the limitation in the header

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+    SimpleBranchPredictor(uint32_t max_fetch_insts) :
+        max_fetch_insts_(max_fetch_insts)
+    {}
+    DefaultPrediction getPrediction(const DefaultInput &);
+    void updatePredictor(DefaultUpdate &);
+private:
+    // maximum number of instructions in a FetchPacket
+    uint32_t max_fetch_insts_; 
+    // a map of branch PC to 2 bit staurating counter tracking branch history
+    std::map <uint64_t, uint8_t> branch_history_table_; // BHT
```
rajatbhatia1, line 92: BHT and BTB are not size limited here. This seems not physically realizable in silicon

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+// following class definitions are example inputs & outputs for a very simple branch
+// predictor
+class DefaultPrediction
+{
+public:
+    // index of branch instruction in the fetch packet
+    // branch_idx can vary from 0 to (FETCH_WIDTH - 1)
+    uint32_t branch_idx;
+    // predicted target PC 
+    uint64_t predictedPC;
```
rajatbhatia1, line 49: inconsistent coding style. Camelcase is for function names, not memeber variables
use: predicted_PC

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include <cstdint>
+#include <map>
+#include "sparta/utils/SpartaAssert.hpp"
+
+template <class PredictionT, class UpdateT, class InputT>
+class BranchPredictorIF
+{
+public:
+    virtual PredictionT getPrediction(const InputT &) = 0;
```
rajatbhatia1, line 36: should be a const return type, to allow compiler to do move optimization
use: virtual const PredictionT getPrediction(const InputT &) = 0;

core/BranchPred.cpp
oms-vmicro: I believe this is initialized to 1 to indicate a starting point in the state-machine of "weakly not taken" vs. 0 which is "strongly not taken".

It may be worth explicitly stating the state-machine states in the header description for the implemented "SimpleBranchPredictor", just so it is clear what algorithm is implemented.  Or maybe just link to a description of what algorithm is implemented (e.g. wiki).

core/BranchPred.cpp
```diff
@@ -0,0 +1,70 @@
...
+        branch_history_table_[update.FetchPC] =
+            (branch_history_table_[update.FetchPC] == 0) ? 0 :
+             branch_history_table_[update.FetchPC] - 1;
+    }
+}
+
+DefaultPrediction SimpleBranchPredictor::getPrediction(const DefaultInput & input) {
+    bool predictTaken;
+    if (branch_history_table_.find(input.FetchPC) != branch_history_table_.end()) {
+        predictTaken = (branch_history_table_[input.FetchPC] > 1);
```
oms-vmicro, line 42: A comment noting that history-table states of "3" and "2" as taken vs. "1" and "0" as not-taken would be helpful.  A more detailed header description making note of the implemented algorithm would negate this suggestion to add a comment here.

core/BranchPred.cpp
oms-vmicro: In addition to documenting that this implementation of SimpleBranchPredictor doesn't support RVC extension, would we want to provide guidance on how a user could support RVC via this API?  Would we need to enhance the API in order to support RVC?  Maybe these are questions that can get bumped to a future enhancement.

core/BranchPred.cpp
```diff
@@ -0,0 +1,70 @@
...
+        branch_target_buffer_[update.FetchPC].predictedPC = update.correctedPC;
+    } else {
+        branch_history_table_[update.FetchPC] =
+            (branch_history_table_[update.FetchPC] == 0) ? 0 :
+             branch_history_table_[update.FetchPC] - 1;
+    }
+}
+
+DefaultPrediction SimpleBranchPredictor::getPrediction(const DefaultInput & input) {
+    bool predictTaken;
```
knute-mips, line 40: initialize to `false`

core/BranchPred.cpp
```diff
@@ -0,0 +1,70 @@
...
+ *       - if not present in BTB entry, prediction branch idx is the last instr of
+ *         the FetchPacket, while predicted PC is the fall through addr. Also, create
+ *         a new BTB entry 
+ * Update:
+ *    - a valid BTB entry must be present for fetch PC
+ *    - TBD 
+ *
+ */
+
+#define BYTES_PER_INST 4
```
knute-mips, line 21: This can be a `constexpr` in the `BranchPredictorIF` class

core/BranchPred.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+void SimpleBranchPredictor::updatePredictor(DefaultUpdate & update) {
```
knute-mips, line 23: Does not appear that the `DefaultUpdate` object is modified in this function.  Make it `const`

core/BranchPred.hpp
knute-mips, line 33: Class is missing a `virtual` destructor.

core/BranchPred.hpp
knute-mips: 💯 

Also should initialized to defaults to catch errors:
```
    uint32_t branch_idx = std::numeric_limits<uint32_t>::max();
    // predicted target PC 
    uint64_t predictedPC = std::numeric_limits<uint64_t>::max();
```

core/BranchPred.hpp
```diff
@@ -0,0 +1,96 @@
...
+{
+public:
+    SimpleBranchPredictor(uint32_t max_fetch_insts) :
+        max_fetch_insts_(max_fetch_insts)
+    {}
+    DefaultPrediction getPrediction(const DefaultInput &);
+    void updatePredictor(DefaultUpdate &);
+private:
+    // maximum number of instructions in a FetchPacket
+    uint32_t max_fetch_insts_;
```
knute-mips, line 90: Can be `const`

knute-mips (changes requested):

core/BranchPred.hpp
danbone: Seconded

core/BranchPred.hpp
arupc-vmicro: Yes, but the intent for this simple predictor is to have an infinite size BTB and BHT, such that only misses are cold misses. I will add a comment explaining that.

core/BranchPred.cpp
arupc-vmicro: Expanded comment at the beginning of this file as an explanation.

core/BranchPred.cpp
arupc-vmicro: Added a TODO comment in branch predictor IF class definition

arupc-vmicro:
> Please put everything in separate namespace

Have added branch prediction IF and SimplePredictor in the Olympia namespace, for now, in line with what we have currently for LSU, Dispatch etc. I will capture this request in a separate issue to create separate namespaces for different units of Olympia

arupc-vmicro:
> On a different note do we need to discuss the directory structure going forward for when others implement their own predictors? Do we need a separate branchpred directory under core?

Let us address this in a new PR when we add other branch predictors.

core/BranchPred.hpp
arupc-vmicro: Can you please explain further why this is needed / recommended ? 

This https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f49-dont-return-const-t ,  seem to suggest that const return type would suppress use of move semantics and is not advisable.

arupc-vmicro:
Addressed review comments. Please take a look and approve, if looks good. I need at least one approval for merge.

danbone (changes requested):

core/SimpleBranchPred.hpp
```diff
@@ -0,0 +1,86 @@
...
+    class DefaultUpdate
+    {
+    public:
+        uint64_t fetch_PC = std::numeric_limits<uint64_t>::max();
+        uint32_t branch_idx = std::numeric_limits<uint32_t>::max();
+        uint64_t corrected_PC = std::numeric_limits<uint64_t>::max();
+        bool actually_taken = false;
+    };
+
+    class DefaultInput
```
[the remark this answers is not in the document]
arupc-vmicro: Added a separate namespace for BranchPredictor within olympia.

core/BranchPredIF.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace olympia
+{
+
+    template <class PredictionT, class UpdateT, class InputT>
+    class BranchPredictorIF
+    {
+    public:
+        // TODO: create constexpr for bytes per compressed and uncompressed inst
+        static constexpr uint8_t bytes_per_inst = 4;
```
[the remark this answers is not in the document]
arupc-vmicro: `bytes_per_inst` being a property of RV32 ISA, it is part of BranchPredictorIF and not part of SimpleBranchPredictor, which is a specific implementation of the predictor

arupc-vmicro:
As a reminder, if there is no further comment, please approve so that we can merge this.

danbone (approved):

arupc (approved):

PR #145, opened 2024-01-27
Add an example of PEvent generation

klingaard:
This SHA is an example of how to generate PEvents (Performance Events) typically used for correlation 
efforts between a perf model and hardware.

PEvents are a more "compact" form of the pipeout collection pairs, allowing a modeler to specify more direct items 
for collection.  Pipeout pairs can be more elaborate/detailed.  However, with some clever coding, there can be
reuse -- it's not done here.

To generate performance events, use the command line parameter `--pevents <filename> <pevent1,pevent2,...|all>`.

Examples:
```
# All pevents
./olympia traces/dhry_riscv.zstf --pevents retire.out all -i100

# Just RETIRE pevent
./olympia traces/dhry_riscv.zstf --pevents retire.out RETIRE -i100

# RETIRE and COMPLETE pevent
./olympia traces/dhry_riscv.zstf --pevents retire.out COMPLETE,RETIRE -i100
```

While I was at it, I removed the generated annotation for pipeline collection for LoadStore/MemoryAccess classes -- uses NVP which generates faster, as well as creates smaller pipeouts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,39 @@ dyff between baseline.yaml always_hit_DL1.yaml
    -l top.*.*.decode info decode_rob.log \
    -l top.*.*.rob    info decode_rob.log
 ```
+### Generate PEvents (for Correlation)
+
+PEvents or Performance Events are part of the Sparta Modeling
+Framework typically used to correlate a performance model with RTL.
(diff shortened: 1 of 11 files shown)


furuame (approved):
Looks good to me! You may also add a test for it.

PR #146, opened 2024-01-29
fusion PR, removed DSL

jeffnye-gh:
This isolates the Fusion API from the domain specific language.

After an agreement on syntax the DSL would be added as an improvement to the API.


fusion/fusion/FieldExtractor.hpp
```diff
@@ -0,0 +1,279 @@
...
+        }
+        return value;
+        //        return getSFieldById(inst, field);
+    }
+
+    //! \brief get the encoded value of the full immediate field
+    //!
+    //! split immediate fields are (will be) ordered msb:lsb and
+    //! concatenated into one unsigned value
+    uint32_t getImmField(InstPtrType inst) const { return 0; }
```
rajatbhatia1, line 101: Is this not implemented? If so, should we throw an exception till it is implemented?

fusion/fusion/FieldExtractor.hpp
```diff
@@ -0,0 +1,279 @@
...
+        FieldName f2 = _f2.value_or(f1);
+
+        auto lhs = getField(LHS, f1);
+        auto rhs = getField(RHS, f2);
+
+        switch (func)
+        {
+            case FUNC::LT: { return lhs < rhs;  }
+            case FUNC::EQ: { return lhs == rhs; }
+            default:       { return false;      }
```
rajatbhatia1, line 263: default case should be assert, as we should never call compare with anything other than LT/EQ. It is a coding error.

fusion/fusion/FieldExtractor.hpp
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
```
rajatbhatia1, line 265: I am not sure if latest compiler versions have fixed this, but older compiler versions were not smart enough to detect that this function will indeed return a bool in all cases.  Suggest moving the return false here

fusion/fusion/FieldExtractor.hpp
jeffnye-gh: Yes this is not implemented. I intended to remove this and mark it as a todo.

fusion/fusion/FieldExtractor.hpp
jeffnye-gh: ok

fusion/fusion/FieldExtractor.hpp
jeffnye-gh: ok simple change.

jeffnye-gh: neither LLVM or GCC complains. but the change is easy enough.

fusion/fusion/FusionContext.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+      public:
+        FusionContext() = default;
+
+        //! \brief basic ctor
+        //!
+        //! name_ member is assigned in makeContext
+        FusionContext(const std::string & name, const FusionGroupListType & groups)
+        {
+            makeContext(name_, groups);
```
oms-vmicro, line 49: Typo?  Should this be `name` instead of `name_` for the first parameter?

fusion/fusion/FusionGroup.hpp
```diff
@@ -0,0 +1,303 @@
...
+            // std::cout<<"  ASM: ";
+            // for(auto a : asms_) std::cout<<a<<std::endl;
+        }
+
+      protected:
+        //! \brief Method to calculate hash based on UIDs
+        //!
+        //! In future consider making this a user controlled functor.
+        //! In hardware the adds are not desirable.
+        static HashType jenkins_1aat(const std::vector<UidType> & v)
```
oms-vmicro, line 160: Realize this may get replaced with a user-controlled functor in the future and be moot, but was curious if it would be worth adding a comment as to the logic behind the implemented hash function.  Maybe similar to the comments regarding the RadixTrie (which I appreciated!).

fusion/fusion/FusionContext.hpp
jeffnye-gh: yes that is a typo, corrected.

fusion/fusion/FusionGroup.hpp
jeffnye-gh: I have added some additional description.

fusion/fusion/FusionContext.hpp
```diff
@@ -0,0 +1,94 @@
...
+                     const FusionGroupListType & groups)
+    {
+        name_ = n;
+        for (const auto & grp : groups)
+        {
+            insertGroup(grp);
+        }
+    }
+
+    //! \brief insert each group w/ catch
```
rajatbhatia1: Just a nit
//! \brief insert each group in current context w/ catch

fusion/fusion/FusionContext.hpp
```diff
@@ -0,0 +1,94 @@
...
+        }
+        container_.insert(std::make_pair(hash, group));
+    }
+
+  private:
+    //! \brief context name
+    //!
+    //! When this is used in (future) multi-context implementations
+    //! the name must be unique
+    std::string name_{""};
```
rajatbhatia1: let's set a default name, so the exceptions strings (and log messages) have same string pattern. Will make log parsing and error checking easier in future.

fusion/test/TestBench.hpp
```diff
@@ -0,0 +1,254 @@
...
+//!
+//! There is a callback for each fusion group test case  f1, f1.1, etc
+// ---------------------------------------------------------------------
+struct TestBench::cbProxy
+{
+    //! \brief ...
+    static bool uf1_func(FusionGroupType &, InstPtrListType &,
+                         InstPtrListType &)
+    {
+        cout << "HERE uf1_func called" << endl;
```
rajatbhatia1, line 164: shouldn't we be suing Msg class instead of couts?

fusion/fusion/FusionContext.hpp
jeffnye-gh: done

klingaard (approved):

klingaard:
@jeffnye-gh, looks like 1 conversation yet to be resolved.  Are you ready for me to push the big green button?

fusion/test/TestBench.hpp
jeffnye-gh: fixed

jeffnye-gh:
sorry, yes just resolved the conversation, the changes are in.

jeffnye-gh:
yes, the changes were in, i did not close the comment. Just closed the
conversation.

PR #148, opened 2024-01-31
Updated to latest version of the STF lib submodule

kathlenemagnus:
Improvements to Dromajo's STF tracing support require the latest version of the STF library.


klingaard (approved):

klingaard:
Is this backward compatible with existing traces or do they need to be regenerated?

kathlenemagnus:
> Is this backward compatible with existing traces or do they need to be regenerated?

It's expected to be backwards compatible. The fix I needed to pull in allows a "force PC record" to be added at anytime, which is needed to support non-contiguous traces.

klingaard:
Got it.  Thanks!

PR #151, opened 2024-02-08
Sparta Assert for Target Pipe without Execution Unit

AaronGChan:
Forgot to add an assert to check that there is a dispatcher (thus a subsequent issue queue/execution unit) that can handle a target pipe of an instruction. If one forgets to define one, i.e you forget to define an execution unit to handle `DIV` target pipe, the simulator will end early, but no error will be thrown.

### DIFF
--- a/core/Dispatch.cpp
+++ b/core/Dispatch.cpp
@@ -243,10 +243,10 @@ namespace olympia
                               "Pipe Target: "
                                   << target_pipe
                                   << " doesn't have any execution unit that can handle that target "
-                                     "pipe. Did you define it in the yaml properly?")
-                    // so we have a map here that checks for which valid dispatchers for that
-                    // instruction target pipe map needs to be: "int": [exe0, exe1, exe2]
-                    if (target_pipe != InstArchInfo::TargetPipe::LSU)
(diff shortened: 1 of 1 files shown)


core/Dispatch.cpp
```diff
@@ -239,9 +239,14 @@ namespace olympia
...
-                // instruction target pipe map needs to be: "int": [exe0, exe1, exe2]
-                if (target_pipe != InstArchInfo::TargetPipe::LSU)
+                sparta_assert(dispatchers.size() > 0,
+                              "Pipe Target: "
+                                  << target_pipe
+                                  << " doesn't have any execution unit that can handle that target "
+                                     "pipe. Did you define it in the yaml properly?")
+                    // so we have a map here that checks for which valid dispatchers for that
+                    // instruction target pipe map needs to be: "int": [exe0, exe1, exe2]
+                    if (target_pipe != InstArchInfo::TargetPipe::LSU)
```
klingaard, line 249: Indention is funny.   I think it's probably the missing semicolon on the assertion

AaronGChan: Added semicolon, didn't seem to fix formatting too much, this is with clang-format too.

PR #154, opened 2024-02-08
Random branch misprediction flush fix

### LINKED ISSUE #153: Random branch misprediction/flushing has a race condition
Flushing from the ExecutionPipe, unfortunately, does not work.  There is a race condition where ROB could retire a flushing branch _before_ the flush manager has a chance to flush it.  I have a patch for this.

klingaard:
CC: @danbone 

closes #153

### DIFF
--- a/core/ROB.cpp
+++ b/core/ROB.cpp
@@ -171,6 +171,14 @@ namespace olympia
                     break;
                 }
 
+                // Is this a misprdicted branch requiring a refetch?
+                if(ex_inst.isMispredicted()) {
+                    FlushManager::FlushingCriteria criteria(FlushManager::FlushCause::MISPREDICTION,
+                                                            ex_inst_ptr);
+                    out_retire_flush_.send(criteria);
+                    break;
+                }
+
                 // This is rare for the example
                 if(SPARTA_EXPECT_FALSE(ex_inst.getPipe() == InstArchInfo::TargetPipe::SYS))
                 {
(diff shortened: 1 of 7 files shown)


danbone:
This is fine, though it removes the ability to redirect mispredicted branches as soon as they're executed (waiting for them to reach the head of the ROB instead).

core/ROB.cpp
```diff
@@ -171,6 +171,14 @@ namespace olympia
...
                 }
 
+                // Is this a misprdicted branch requiring a refetch?
+                if(ex_inst.isMispredicted()) {
+                    FlushManager::FlushingCriteria criteria(FlushManager::FlushCause::MISPREDICTION,
+                                                            ex_inst_ptr);
+                    out_retire_flush_.send(criteria);
+                    break;
+                }
+
```
danbone, line 181: This is going to hit a similar problem right?

You're relying on the flushmanager to flush the ROB before retireInstructions_ is called again?

klingaard: No, the loop breaks right after detecting the mispredicted load.  Another retiring event is scheduled for next cycle which will be called _after_ the flush occurs.

danbone: It suffers from the same frailties as the previous way right?

You're just moving the race condition from the executepipe to the ROB?

klingaard: You are correct there still is a race condition.  I've added the required delay by breaking the loop and expecting a flush the next cycle.  I'll add code to ensure retirement "stops" until the flush is received.

danbone (approved):

AaronGChan (approved):

PR #160, opened 2024-03-05
FSL.md circulated for review

jeffnye-gh:
FSL.md contains the markdown for the FSL
language description.

Circulating for review comments to refine the syntax.

See the TODO section, in-place vs copy-on-write for instruction buffers and value of SA in the syntax.

### DIFF
--- a/fusion/FSL.md
+++ b/fusion/FSL.md
@@ -1,0 +1,1197 @@
+# Fusion/Fracture Specification Language 
+
+1. Document History
+
+1. Introduction to FSL
+    1. High Level Operation
+    1. FSL Example
+    1. Is/Is not
+
+1. Language Description
+    1. Scope
+    1. Keywords
+    1. Identifiers
+    1. Constants
+    1. Operators
+        1. Arithmetic Operators
+        1. Range Operators
+        1. Relational Operators
+        1. Logical Operators
+        1. Assignment Operators
+        1. Concatenation Operator
+    1. Comments
+    1. Preprocessor Support
+        1. Conditional Compilation
+        1. Includes
+    1. Transform Structure
+    1. Transform Specification
+    1. Prolog Elements
+    1. isa Element
+        1. isa Methods
+    1. uarch Element
+        1. uarch methods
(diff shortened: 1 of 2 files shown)


arupc-vmicro:
@rajatbhatia1 @knute-mips Can you please take a look at this PR?

fusion/FSL.md
```diff
@@ -0,0 +1,1206 @@
...
+expense to performance.
+
+The example sequences above have an implied strict ordering. There are 
+sequence pragmas to relax the strictness of sequence matching. 
+
+The required pragma, \_req\_, indicates that an unspecified instruction
+is required in that position. The \_req\_ case does not constrain 
+what instruction can be present in the gap.
+
+The optional pragma, \_opt\_, indicates that the match will not be
```
rajatbhatia1, line 685: Is there a need for this opt instruction to possibly be a variable number of instructions (e.g. 1-3 optional instructions)

jeffnye-gh: I will add an integer to the syntax, __opt__ 3, meaning any of 1,2,3 optional instructions. 

Implication is that no instructions in the range is also legal, since each optional instruction is 0 or 1.

in the syntax __opt__ will be treated the same as __opt__ 1

Makes sense to do the same for __req__ 3

Thanks.

jeffnye-gh: @rajatbhatia1 I have made these changes they are included in [PR #163](https://github.com/riscv-software-src/riscv-perf-model/pull/163). I am going to close this PR it is redundant now.
```
The required pragma, \_req\_, indicates that an unspecified instruction
is required in that position. The \_req\_ case does not constrain
what instruction can be present in the gap. A trailing integer
can specify more than one required instruction.

The optional pragma, \_opt\_, indicates that the match will not be
rejected if there is an instruction or not in this position. A
trailing integer can be used to specify up to N optional instructions.

_opt_ 2    indicates 0 - 2 instructions will match
_req_ 2    indicates there must be 2 instructions in this position
           in the sequence.
```

PR #166, opened 2024-04-22
Non blocking cache implementation

h0lyalg0rithm:
The implementation is based on the previous work by Vineeth.
However I made change to use a single pipeline and reduce the number of events generated inside the Dache.
I has also removed the col-easing of the mshr requests (based on the block address) instead each request generates an mshr and the same instruction is then sent to next cache for lookup in case of a miss. 

Micro architecture details https://docs.google.com/document/d/1HLlCkfZUtt6BafgVypS5pwS90zo4XOrGFOR1KIzYHLw/edit?usp=sharing


core/DCache.cpp
```diff
@@ -35,71 +66,246 @@ namespace olympia {
...
+        }
+
+        // Enqueue Load in LMQ
+        ILOG("Load miss inst to LMQ; block address:0x" << std::hex << block_addr);
+        (*mshr_it)->setMemRequest(mem_access_info_ptr);
+        mem_access_info_ptr->setCacheState(MemoryAccessInfo::CacheState::MISS);
+        out_lsu_lookup_req_.send(mem_access_info_ptr);
+    }
+
+    uint64_t DCache::getBlockAddr(const MemoryAccessInfoPtr & mem_access_info_ptr)
```
klingaard, line 191: This method can be `const`

core/DCache.hpp
```diff
@@ -26,29 +28,77 @@ namespace olympia
...
         bool dataLookup_(const MemoryAccessInfoPtr & mem_access_info_ptr);
 
         void reloadCache_(uint64_t phy_addr);
 
-        void getInstsFromLSU_(const MemoryAccessInfoPtr & memory_access_info_ptr);
+        uint64_t getBlockAddr(const MemoryAccessInfoPtr & mem_access_info_ptr);
+
+        // To arbitrate between incoming request from LSU and Cache refills from BIU
+        //        bool incoming_cache_refill_ = false;
+        MemoryAccessInfoPtr incoming_cache_refill_ = nullptr;
```
klingaard, line 62: Instead of "remembering" the incoming cache refill, why not have the MemoryAccessInfoPtr class indicate that it's a refill?

core/DCache.cpp
```diff
@@ -35,71 +66,246 @@ namespace olympia {
...
+        return addr_decoder_->calcBlockAddr(inst_target_addr);
+    }
+
+    // Data read stage
+    void DCache::handleDataRead_()
+    {
+        const auto stage_id = static_cast<uint32_t>(PipelineStage::DATA_READ);
+        const MemoryAccessInfoPtr & mem_access_info_ptr = cache_pipeline_[stage_id];
+        ILOG(mem_access_info_ptr << " in read stage");
+        if (incoming_cache_refill_ == mem_access_info_ptr)
```
klingaard, line 204: If the MemoryAccessInfo keeps track of whether it's a refill or not, this code can change to:
```
if(mem_access_info_ptr->isRefill())
```

core/DCache.hpp
```diff
@@ -91,6 +145,13 @@ namespace olympia
         sparta::Counter dl1_cache_misses_{getStatisticSet(), "dl1_cache_misses",
                                           "Number of DL1 cache misses",
                                           sparta::Counter::COUNT_NORMAL};
+
+
+        sparta::Buffer<MSHREntryInfoPtr> mshr_file_;
+        MSHREntryInfoAllocator & mshr_entry_allocator;
+        void allocateMSHREntry_(const MemoryAccessInfoPtr & mem_access_info_ptr);
+        void mshrLookup_(const MemoryAccessInfoPtr & mem_access_info_ptr);
+        void replyLSU(const MemoryAccessInfoPtr & mem_access_info_ptr);
```
klingaard, line 154: Is this method private?  Missing trailing underscore...

core/DCache.cpp
```diff
@@ -35,71 +66,246 @@ namespace olympia {
...
+    }
+
+    void DCache::getMemReqFromLSU_(const MemoryAccessInfoPtr & memory_access_info_ptr)
+    {
+
+        ILOG("Got memory access request from LSU " << memory_access_info_ptr);
+        if (!cache_refill_selected_)
+        {
+            ILOG("Arbitration from refill " << memory_access_info_ptr);
+            memory_access_info_ptr->setCacheState(MemoryAccessInfo::CacheState::MISS);
```
klingaard, line 284: It's not really a miss, but a loss of arbitration.  Does the LSU know to try again next cycle?

klingaard (commented):
Couple of things to note:
* There's no guarantee that a response from the L2 cache (i.e. a refill) will take precedence over a new request from the LS.  To fix this, you might need to add a precedence rule between the L2 response port and the LSU request
* If the L2 sends a refill response for cycle N, you obviously need to block the LS from arbitrating.  You've done that with the boolean `cache_refill_selected_`.  But you can also do that but querying the pipeline "have you been appended.."  This removes the need for the extra event to clear the new boolean.

core/DCache.hpp
```diff
@@ -26,29 +28,77 @@ namespace olympia
...
             PARAMETER(bool, l1_always_hit, false, "DL1 will always hit")
+            PARAMETER(uint32_t, mshr_entries, 8, "Number of MSHR Entries")
         };
 
         static const char name[];
         DCache(sparta::TreeNode* n, const CacheParameterSet* p);
 
       private:
+        ////////////////////////////////////////////////////////////////////////////////
+        // L1 Cache Handling
```
jhsiaovt, line 39: Since this is DCache, would you change all references to L1 Cache to DCache unless you already have some other plans for ICache/IL1Cache.

core/DCache.hpp
```diff
@@ -61,8 +111,8 @@ namespace olympia
 
         sparta::DataInPort<uint32_t> in_l2cache_ack_{&unit_port_set_, "in_l2cache_ack", 1};
```
jhsiaovt, line 112: Add comment this port is receiving an ack from next level memory. If there's any possibility of connecting DCache to L3 or BIU directly, then this method name becomes confusing; consider changing method name if possibility exist.

core/DCache.hpp
```diff
@@ -61,8 +111,8 @@ namespace olympia
...
 
-        sparta::DataInPort<MemoryAccessInfoPtr> in_l2cache_resp_{&unit_port_set_,
-                                                                 "in_l2cache_resp", 1};
+        sparta::DataInPort<MemoryAccessInfoPtr> in_l2cache_resp_{&unit_port_set_, "in_l2cache_resp",
+                                                                 1};
```
jhsiaovt, line 115: Add comment as to what "resp" is - ack or fill data or something else.

core/DCache.hpp
```diff
@@ -91,6 +145,13 @@ namespace olympia
         sparta::Counter dl1_cache_misses_{getStatisticSet(), "dl1_cache_misses",
                                           "Number of DL1 cache misses",
                                           sparta::Counter::COUNT_NORMAL};
+
```
jhsiaovt, line 148: If you want to be nice to the end user, add a hit and/or a miss ratio.  ;)

core/DCache.cpp
```diff
@@ -35,71 +66,246 @@ namespace olympia {
...
+            mem_access_info_ptr->setCacheState(MemoryAccessInfo::CacheState::HIT);
+            out_lsu_lookup_ack_.send(mem_access_info_ptr);
+            return;
+        }
+
+        // Enqueue Load in LMQ
+        ILOG("Load miss inst to LMQ; block address:0x" << std::hex << block_addr);
+        (*mshr_it)->setMemRequest(mem_access_info_ptr);
+        mem_access_info_ptr->setCacheState(MemoryAccessInfo::CacheState::MISS);
+        out_lsu_lookup_req_.send(mem_access_info_ptr);
```
jhsiaovt, line 188: Move out_lsu_lookup_req_.send(...) out of both if statements. Have only one call to out port. This greatly simplifies debug.
Remove the return statements. Also simplifies debug.
Change second "if" to "else if"
Add "else" to enqueue load.
All in the name of debugging.

core/DCache.hpp
```diff
@@ -26,29 +28,77 @@ namespace olympia
...
+        ////////////////////////////////////////////////////////////////////////////////
+        // Handle requests
+        ////////////////////////////////////////////////////////////////////////////////
+
+        void getMemReqFromLSU_(const MemoryAccessInfoPtr & memory_access_info_ptr);
 
         void getAckFromL2Cache_(const uint32_t & ack);
 
         void getRespFromL2Cache_(const MemoryAccessInfoPtr & memory_access_info_ptr);
```
jhsiaovt, line 93: Comment within body for these get* methods says "Got..."; consider renaming to "receive". "get" also implies active requester whereas "receive" implies passive acceptor.

core/MSHREntryInfo.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+        void setValid(bool v) { line_fill_buffer_.setValid(v); }
+
+        bool isModified() const { return line_fill_buffer_.isModified(); }
+
+        void setModified(bool m) { line_fill_buffer_.setModified(m); }
+
+        void setDataArrived(bool v) { data_arrived_ = v; }
+
+        bool getDataArrived() { return data_arrived_; }
```
jhsiaovt, line 30: Change name to isDataArrived() to match bool return type.

core/DCache.cpp
h0lyalg0rithm: since I am returning an int64 I think the const is ignored and I get the following compiler error
`error: type qualifiers ignored on function return type [-Werror=ignored-qualifiers]`

core/DCache.cpp
h0lyalg0rithm: I have replaced it with RELOAD as i think its also used in the l2 cache for a similar situation

core/DCache.cpp
h0lyalg0rithm: The function is const and not the return type. will make the changes

core/DCache.cpp
```diff
@@ -20,12 +20,16 @@ namespace olympia
...
+        in_l2cache_resp_.registerConsumerHandler(
+            CREATE_SPARTA_HANDLER_WITH_DATA(DCache, receiveRespFromL2Cache_, MemoryAccessInfoPtr));
+
         in_l2cache_ack_.registerConsumerHandler(
             CREATE_SPARTA_HANDLER_WITH_DATA(DCache, receiveAckFromL2Cache_, uint32_t));
 
-        in_l2cache_resp_.registerConsumerHandler(
-            CREATE_SPARTA_HANDLER_WITH_DATA(DCache, receiveRespFromL2Cache_, MemoryAccessInfoPtr));
+        in_l2cache_resp_.registerConsumerEvent(in_l2_cache_resp_receive_event_);
+        in_lsu_lookup_req_.registerConsumerEvent(in_lsu_lookup_req_receive_event_);
```
h0lyalg0rithm, line 30: @knute-mips I was not 100% if this is what you meant in the last call.

core/DCache.hpp
```diff
@@ -132,6 +132,17 @@ namespace olympia
 
         sparta::UniqueEvent<> uev_mshr_request_{
             &unit_event_set_, "mshr_request", CREATE_SPARTA_HANDLER(DCache, mshrRequest_)};
+
+        void noOpEventHandler() {}
```
h0lyalg0rithm, line 136: @knute-mips Is there a way to setup the unique event without this handler

core/DCache.cpp
klingaard: Yes!  Actually it is.  Sorry, meant to give you an example of it.  Glad you figured it out!

h0lyalg0rithm:
@klingaard @jhsiaovt The PR is ready to review, I have added the test case to validate the arbitration logic

core/DCache.hpp
```diff
@@ -81,6 +121,38 @@ namespace olympia
         ////////////////////////////////////////////////////////////////////////////////
         // Events
         ////////////////////////////////////////////////////////////////////////////////
+        sparta::UniqueEvent<> uev_mshr_request_{
+            &unit_event_set_, "mshr_request", CREATE_SPARTA_HANDLER(DCache, mshrRequest_)};
+
+
+        sparta::utils::ValidValue<MemoryAccessInfoPtr> l2_mem_access_info_;
+        sparta::utils::ValidValue<MemoryAccessInfoPtr> lsu_mem_access_info_;
+        void arbitrate_l2_lsu_req_()
```
klingaard, line 130: Can you change the function name to match the coding style?

test/core/dcache/next_lvl_cache_refill.json
klingaard, line 1: I think `lw` instructions only have 1 RS

klingaard (approved):

core/DCache.hpp
h0lyalg0rithm: My bad I should have fixed it before, I will create a PR to add this linting style to the github actions.

jhsiaovt (approved):
Thank you for the changes and adding a test suite!

h0lyalg0rithm:
> @h0lyalg0rithm are you good to merge this? I need to account for your port changes in my VLSU PR.

@aarongchan Yes I am ready to merge this

PR #167, opened 2024-05-06
Vector 1.0 Support VALU

AaronGChan:
PR Goals:
Implement basic vector support for VALU operations, UOp generation for LMUL > 1, and vset support. The goal in this stage to design all necessary components to support a vadd.vv instruction with varying LMUL and SEW values set.

Documentation:
https://github.com/riscv-software-src/riscv-perf-model/discussions/89#discussioncomment-9324963


core/CPUTopology.cpp
```diff
@@ -395,6 +397,16 @@ void olympia::CoreTopologySimple::bindTree(sparta::RootTreeNode* root_node)
...
                     core_node + ".execute." + unit_name + ".ports.out_execute_pipe";
                 bind_ports(exe_pipe_in, exe_pipe_out);
+
+                for(uint32_t i = 0; i < pipelines[pipe_idx].size(); ++i){
+                    if(pipelines[pipe_idx][i] == "vset"){
+                        // only bind execute pipe -> decode port for an issue queue if it has a vset pipe
+                        const std::string exe_vset_out =
+                            core_node + ".execute." + unit_name + ".ports.out_vset";
+                        bind_ports(vset_in_decode, exe_vset_out);
+                        break; // break after because there should only be one vset per issue queue
```
rajatbhatia1, line 407: Should we have a check and assert instead of break to catch possible misconfiguration issues?

core/Decode.cpp
```diff
@@ -148,20 +172,125 @@ namespace olympia
...
+                    // mavis_id 216 -> vsetivli
+                    // TODO: change back from getMnemonic() to getInstructionUniqueID() where the ID doesn't change
+                    if(inst->getOpCodeInfo()->getInstructionUniqueID() == mavis_vsetivli_uid_){
+                        // vsetivli with immediates, we can set at decode and continue to process instruction group, no vset stall
+                        VCSRs_.lmul = inst->getLMUL();
+                        VCSRs_.vl = inst->getVL();
+                        VCSRs_.vta = inst->getVTA();
+                        VCSRs_.sew = inst->getSEW();
+                    }
+                    else if(inst->isVset() && (inst->getSourceOpInfoList()[0].field_value != 0 || inst->getOpCodeInfo()->getInstructionUniqueID() == mavis_vsetvl_uid_))
```
rajatbhatia1, line 199: please break statement into multiple lines for legibility

AaronGChan: Clang format applied in [b8028a3](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/b8028a38dd6f24d24687a0984feca2d5f382d1a9)

core/CPUTopology.cpp
AaronGChan: Added sparta assertion checking for duplicate pipe definitions in [b8028a3](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/b8028a38dd6f24d24687a0984feca2d5f382d1a9)

core/Decode.hpp
```diff
@@ -120,6 +123,15 @@ namespace olympia
             //! \brief ...
             PARAMETER(FileNameListType, fusion_group_definitions, {},
                       "Lists of fusion group UID json files")
+
+            //! LMUL
```
kathlenemagnus, line 127: Since we allow vset instructions to change these values, can we rename these params to something like `init_lmul`, `init_sew` and `init_vl`?

core/Decode.cpp
```diff
@@ -60,20 +61,30 @@ namespace olympia
...
     // Send fetch the initial credit count
-    void Decode::sendInitialCredits_() { fetch_queue_credits_outp_.send(fetch_queue_.capacity()); }
+    void Decode::sendInitialCredits_()
+    {
+        fetch_queue_credits_outp_.send(fetch_queue_.capacity());
+
+        // setting MavisIDs for vsetvl and vsetivli
+        mavis_facade_ = getMavis(getContainer());
+        mavis_vsetvl_uid_ = mavis_facade_->lookupInstructionUniqueID("vsetvl");
+        mavis_vsetivli_uid_ = mavis_facade_->lookupInstructionUniqueID("vsetivli");
```
kathlenemagnus, line 78: Do we support `vsetvli`?

core/Decode.cpp
```diff
@@ -112,6 +123,20 @@ namespace olympia
...
 
+    // process vset settings being forward from execution pipe
+    // for set instructions that depend on register
+    void Decode::process_vset_(const InstPtr & inst)
+    {
+        VCSRs_.lmul = inst->getLMUL();
+        VCSRs_.sew = inst->getSEW();
+        VCSRs_.vl = inst->getVL();
+        VCSRs_.vta = inst->getVTA();
+        VCSRs_.vlmax = (Inst::VLEN / VCSRs_.sew) * VCSRs_.lmul;
```
kathlenemagnus, line 134: Does `vlmax` get initialized properly at the start of simulation with the initial values?

kathlenemagnus: Just answered my own question by looking at `Inst::VCSRs` struct. I would add a method to this struct so the vlmax calculation isn't defined in two places.

core/Decode.cpp
```diff
@@ -112,6 +123,20 @@ namespace olympia
...
+        waiting_on_vset_ = false;
```
kathlenemagnus, line 135: Might want to try only scheduling Decode if `waiting_on_vset_` was true.

core/Decode.cpp
```diff
@@ -148,20 +173,151 @@ namespace olympia
...
+
+                    // mavis_id 214 -> vsetvl
+                    // mavis_id 216 -> vsetivli
+                    // TODO: change back from getMnemonic() to getInstructionUniqueID() where the ID
+                    // doesn't change
+                    if (inst->getOpCodeInfo()->getInstructionUniqueID() == mavis_vsetivli_uid_)
+                    {
+                        // vsetivli with immediates, we can set at decode and continue to process
+                        // instruction group, no vset stall
+                        VCSRs_.lmul = inst->getLMUL();
```
kathlenemagnus, line 201: How about a method to the `VCSRs_` struct called `update` that takes an instruction and updates all of these values? Would reduced code duplication here and in the `process_vset_` method.

kathlenemagnus: These if-statements need to be restructured. I don't think the `vsetvli` when the source is not x0 case is handled. Something like this should work:
```
if(vset) {
    if(non-stalling vset) {
        // process vset
    }
    else {
        // handle stall
    }
}
elseif (vector inst) {
    // uop gen
}

// send to decode
```

core/CoreTypes.hpp
```diff
@@ -9,29 +9,23 @@ namespace olympia::core_types
...
-        "float"
-    };
+    static inline const char* const regfile_names[] = {"integer", "float", "vector"};
 
-    static inline const char * const issue_queue_types[] = {
-        "alu",
-        "fpu",
-        "br"
-    };
+    static inline const char* const issue_queue_types[] = {"alu", "fpu", "br", "vint", "vset"};
```
kathlenemagnus, line 23: Since we support multiple types of vector pipes (`vint`, `vmul`, `vdiv`, etc.) I would change `vint` to just `vec`.

core/ExecutePipe.cpp
```diff
@@ -53,12 +53,52 @@ namespace olympia
...
+            // have to factor in vlen, sew, valu length to calculate how many passes are needed
+            // i.e if VL = 256 and SEW = 8, but our VALU only has 8 64 bit adders, it will take 4
+            // passes to execute the entire instruction if we have an 8 bit number, the 64 bit adder
+            // will truncate, but we have each adder support the largest SEW possible
+            if (ex_inst->getPipe() == InstArchInfo::TargetPipe::VINT)
+            {
+                if (num_passes_needed_ == 0)
+                {
+                    const uint32_t num_passes =
+                        std::ceil((ex_inst->getVL() / ex_inst->getSEW()) / valu_adder_num_);
```
kathlenemagnus, line 78: I think you mean VLEN here instead of VL. VL (vector length) is the number of elements that an instruction can process whereas VLEN is the size of the vector registers. So for example, is VLEN is 128 and SEW is 8 than a single vector uop contains 16 elements. For LMUL 1, the VL would be 16 but increasing the LMUL will multiply the VL. I think this code works, but the method names are confusing.

core/Fetch.cpp
```diff
@@ -63,14 +63,16 @@ namespace olympia
...
             InstPtr ex_inst = inst_generator_->getNextInst(my_clk_);
-            if(SPARTA_EXPECT_TRUE(nullptr != ex_inst))
-            {
+            if(SPARTA_EXPECT_TRUE(nullptr != ex_inst)){
                 ex_inst->setSpeculative(speculative_path_);
                 insts_to_send->emplace_back(ex_inst);
 
                 ILOG("Sending: " << ex_inst << " down the pipe");
+                // we break after vset to break up instruction group
+                if(ex_inst->isVset()) { break; }
```
kathlenemagnus, line 72: Why do we need this? Looks like you have stalls working in Decode so can this be removed?

core/Inst.hpp
```diff
@@ -132,52 +140,30 @@ namespace olympia
             }
         }
 
-        const Status & getStatus() const
-        {
-            return status_state_;
-        }
+        const Status & getStatus() const { return status_state_; }
```
kathlenemagnus, line 143: I prefer these single-line methods. Much easier to read!

core/Inst.hpp
```diff
@@ -362,8 +422,9 @@ namespace olympia
         sparta::memory::addr_t target_vaddr_ =
             0; // Instruction's Target PC (for branches, loads/stores)
         bool is_oldest_ = false;
-        uint64_t unique_id_ = 0;      // Supplied by Fetch
-        uint64_t program_id_ = 0;     // Supplied by a trace Reader or execution backend
+        uint64_t unique_id_ = 0;  // Supplied by Fetch
+        uint64_t program_id_ = 0; // Supplied by a trace Reader or execution backend
+        uint64_t uopid_ = 0;      // Set in decode
```
kathlenemagnus, line 427: If you used Sparta's `ValidValue` class here you wouldn't need `bool has_uops_`.

core/Inst.hpp
```diff
@@ -372,8 +433,17 @@ namespace olympia
...
         const bool is_csr_;
+        const bool is_vector_;
         const bool is_return_;
+        bool has_uops_;
+        bool has_tail_; // Does this vector instruction have a tail?
+        uint64_t uop_done_count_ = 1; // start at 1 because the uop count includes the parent instruction
+        uint64_t uop_count_ = 0;
+        VCSRs VCSRs_;
 
+        bool is_blocking_vset_ = false;
```
kathlenemagnus, line 444: Would be good to comment what a blocking vset is.

core/ROB.cpp
```diff
@@ -80,10 +69,20 @@ namespace olympia
...
+        {
+            if (!i->isUOp())
+            {
+                // rob entries only for main instruction, uops aren't added.
+                reorder_buffer_.push(i);
+                ILOG("retire appended: " << i);
+            }
+            else
+            {
+                ILOG("UOp passed in, not adding to ROB " << i);
```
kathlenemagnus, line 84: For sanity, we could assert that the uop's parent instruction is the same as the youngest ROB entry.

core/IssueQueue.cpp
```diff
@@ -101,29 +100,66 @@ namespace olympia
...
-        {
-            // Insert at the end if we are doing in order issue or if the scheduler is
-            // empty
-            ILOG("Sending to issue queue " << ex_inst);
-            // will insert based on if in_order_issue_ is set
-            // if it is, will be first in first out, if not it'll be by age, so by UniqueID (UID)
-            ready_queue_.insert(ex_inst);
-            ev_issue_ready_inst_.schedule(sparta::Clock::Cycle(0));
+        
+        if(srcs.size() > 1 && srcs[0].rf != srcs[1].rf && ex_inst->isVector()){
```
kathlenemagnus, line 104: I would had a method to the instruction class for this check instead of doing it here. Like `isV2S` and `isS2V`.

kathlenemagnus:
This is a lot of really great work @AaronGChan!

core/ExecutePipe.cpp
AaronGChan: Changed to VLEN in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd). Added a conditional that will use VL if it is less the VLEN/SEW.

core/IssueQueue.cpp
AaronGChan: Added in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/ROB.cpp
AaronGChan: Added assertion in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Inst.hpp
AaronGChan: Addressed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Inst.hpp
AaronGChan: Addressed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Fetch.cpp
AaronGChan: Removed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/CoreTypes.hpp
AaronGChan: This was actually dead code, not used anymore since we define issue queue types through `.yaml`. Removed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Decode.cpp
AaronGChan: Fixed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Decode.cpp
AaronGChan: Addressed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Decode.cpp
AaronGChan: Addressed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Decode.cpp
AaronGChan: Added support in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

core/Decode.hpp
AaronGChan: Renamed in [73ca9f4](https://github.com/riscv-software-src/riscv-perf-model/pull/167/commits/73ca9f4adb0bce0b8756a9b1ab3d88bf96c589bd)

kathlenemagnus-mips (approved):

AaronGChan:
@kathlenemagnus-mips I think you need to approve on your other account for me to be able to merge.

kathlenemagnus (approved):

Issue #164, opened 2024-04-08
Proposal to Add SYS Instructions to the Execution Unit


ah-condor:
So, the existing implementation is functionally correct and operates as follows: All SYS instructions bypass the execution step and proceed directly to the ROB. When the instruction is retired, the ROB will flush. If there is a destination register, the ROB will remove it from the scoreboard.

Knute provided a more elegant solution, as follows: Every SYS instruction will go through an execution unit, most likely the ALU unit. This execution unit will remove the destination register from the scoreboard after the instruction is executed. The execution unit will also determine whether retiring the instruction by the ROB will result in a flush. To facilitate this, we need to add a new flag to the INST structure.

PR #170, opened 2024-05-20
Add SYS execute pipe

ah-condor:
This PR is a followup to https://github.com/riscv-software-src/riscv-perf-model/pull/162.

It is also discussed in: https://github.com/riscv-software-src/riscv-perf-model/issues/164
 
This PR changes the way SYS instructions are executed. Previously, SYS instructions did not go through an execution pipe but instead were dispatched to the ROB without delay, and the ROB handled the scoreboard with respect to the destination register. This PR requires the SYS instructions to pass through an execution pipe, thereby obviating any special scoreboard code in the ROB.  With the exception of CSR read instructions, the ROB will flush all SYS instructions.

This feature can be tested by running example_json.json workload, which is part of the make regress suite.  Also, the sys execute pipe was added to 3 config files big_core, medium_core and small_core.

How was this tested:
make regress passes 83/83.

### DIFF
--- a/core/ROB.cpp
+++ b/core/ROB.cpp
@@ -256,7 +256,7 @@ namespace olympia
         // if SYS instr is not a csr instruction flush
         // if SYS instr is a csr but src1 is not x0, flush
         // otherwise it is a csr read, therefore don't flush
-        if (ex_inst->getMnemonic().substr(0,3)!="csr" || srclist.size()!=0)
+        if (ex_inst->isCSR())
         {   // this is the case if csr instr with src != x0
             DLOG("retiring SYS wr instr with src reg " << ex_inst->getMnemonic() );
 
--- a/core/ROB.cpp
+++ b/core/ROB.cpp
@@ -73,7 +73,6 @@ namespace olympia
 
     void ROB::sendInitialCredits_()
     {
-        setupScoreboardView_() ;
         out_reorder_buffer_credits_.send(reorder_buffer_.capacity());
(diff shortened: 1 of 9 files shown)


core/ROB.cpp
```diff
@@ -250,54 +249,26 @@ namespace olympia
...
-            scoreboard_views_[reg_file]->setReady(dest_bits);
-            DLOG("retiring SYS inst dest reg bits: " <<  sparta::printBitSet(dest_bits));
-        }
-
-        FlushManager::FlushingCriteria criteria(FlushManager::FlushCause::POST_SYNC, ex_inst);
+        auto srclist = ex_inst->getRenameData().getSourceList();
+        // if SYS instr is not a csr instruction flush
+        // if SYS instr is a csr but src1 is not x0, flush
+        // otherwise it is a csr read, therefore don't flush
+        if (ex_inst->getMnemonic().substr(0,3)!="csr" || srclist.size()!=0)
```
AaronGChan, line 259: String comparisons are slower, you can use Mavis info to check for csr:
`ex_inst->getOpCodeInfo()->isInstType(mavis::InstMetaData::InstructionTypes::CSR)`

klingaard: @ah-condor -- can you make this change?

klingaard (commented):
Looks good.  Once you remove the string compare that Aaron found, this is good to go.

core/ROB.cpp
ah-condor: Working on it!

klingaard (approved):

PR #175, opened 2024-06-15
Jeffnye gh/fsl submodule

jeffnye-gh:
convert fusion support to sub-module fsl.

remove the old fusion dir. Add a cmake var so fsl uses olympia/mavis when it is instantiated as a submodule.


core/Decode.cpp
```diff
@@ -1,7 +1,7 @@
 // <Decode.cpp> -*- C++ -*-
 
 #include "Decode.hpp"
-#include "fusion/FusionTypes.hpp"
+#include "fsl_api/FusionTypes.h"
```
klingaard, line 4: Curious, why the change in name?

klingaard (approved):

core/Decode.cpp
jeffnye-gh: In the early days fusion was accurate, I have cases where I want to apply the language to fracture, some RV extension and for vector. 

I chose fsl_api and fsl_interp to separate the namespaces. 

It could have been fsl and fsl_interp for the symmetry and to quiet my OCD.

Olympia is the lead use case, so if you prefer I can change fsl_api to fsl.

jeffnye-gh:
I pushed an update to correct the conflicts. re-requesting review

The push did not trigger the git runner, i do not know if it is supposed to for this change.

PR #177, opened 2024-07-03
Added script to auto-generate the rv64v uarch json

kathlenemagnus-mips:
Added a script to auto-generate the rv64v uarch json file using the Mavis isa json. This will help us keep track of which vector instructions still need to be implemented in Olympia. Olympia will throw an exception when trying to construct an `Inst` object for an unsupported instruction. Added a simple test to make sure that Olympia fails as expected.


klingaard (approved):
Awesome!  Thanks!

AaronGChan (approved):
Looks good to me!

PR #179, opened 2024-07-11
Refactored Vector Uop Generation

kathlenemagnus:
This PR lands the groundwork for expanding vector uop generation to support different types of vector instructions. 
- Moved vector uop generation into a new class `VectorUopGenerator`
- Fixed several issues with vset processing in Decode, including adding asserts to sanity the vector config
- Refactored the operand check function in the Issue Queue
- Fixed bug in `Inst::VCSRs` where vlmax was not updated correctly when updating the lmul or sew
- Updated several vector tests to use sane vector configs and added a ROB tester


core/Decode.cpp
```diff
@@ -125,19 +131,38 @@ namespace olympia
...
-    void Decode::process_vset_(const InstPtr & inst)
+    void Decode::updateVcsrs_(const InstPtr & inst)
     {
         VCSRs_.setVCSRs(inst->getVL(), inst->getSEW(), inst->getLMUL(), inst->getVTA());
-        // AVL setting to VLMAX if rs1 = 0 and rd != 0
-        if (inst->getSourceOpInfoList()[0].field_value == 0
-            && inst->getDestOpInfoList()[0].field_value != 0)
+
+        const uint64_t uid = inst->getOpCodeInfo()->getInstructionUniqueID();
+        if ((uid == mavis_vsetvli_uid_) && inst->hasZeroRegSource())
```
klingaard, line 139: Isn't the `hasZeroRegSource` already checked before calling this function?

core/Inst.hpp
```diff
@@ -324,6 +336,24 @@ namespace olympia
...
+        }
+
+        bool hasZeroRegDest() const
+        {
+            return std::any_of(getDestOpInfoList().begin(), getDestOpInfoList().end(),
+                [](const mavis::OperandInfo::Element & elem)
+                {
+                    return elem.field_value == 0;
+                });
+        }
```
klingaard, line 355: Wonder if this could be part of the static information on the instruction?

Perhaps it doesn't matter...

core/IssueQueue.cpp
```diff
@@ -97,70 +97,43 @@ namespace olympia
...
-                ev_issue_ready_inst_.schedule(sparta::Clock::Cycle(0));
-            }
-        }
-        else{
-            if (srcs.size() > 0)
-            {
-                reg_file = srcs[0].rf;
-            }
+        uint32_t ready = 0;
+        for(auto src : srcs)
```
klingaard, line 102: `const &`?

core/Decode.cpp
klingaard, line 1: I might have missed it, but on a flush does the vector uop generator get cleared?

core/Decode.cpp
kathlenemagnus: Only for `vsetivli` and certain `vsetvli` instructions, but it is also used when a `vsetvl` is completed. See line 164.

core/IssueQueue.cpp
kathlenemagnus: Fixed.

core/Decode.cpp
kathlenemagnus: No it doesn't. I'll work on that today.

kathlenemagnus: I added a `handleFlush` method and hooked it up in Decode.

klingaard (approved):

Issue #142, opened 2024-01-24
Add data/instruction prefetcher


klingaard:
Similar to requirements of a branch prediction API, add a prefetching API to Olympia and integrate or implement basic data/instruction prefetchers into the model.

The effort would be:
- [ ] Design and propose a prefetching API
- [ ] Add testing for that API
- [ ] Integrate into existing Ldst/Fetch blocks and test

Shubhf:
Hey @rajatbhatia1 @klingaard Can I take up this issue?

PR #180, opened 2024-07-16
Prefetcher API

rajatbhatia1:
Initial version of prefetcher API
    
    This API proposes to separate the prefetcher interface (defined in PrefetcherIF.hpp) with actual implementation of prefetch engine (defined in PrefetchEngineIF.hpp). This approach allows the model to rapidly customize the prefetchers with multiple prefetch policies
    e.g. We can create stride prefetch engine and use the same engine in L2 Prefetcher as well as L3 prefetcher.
    
    Prefetchers that depend on external (to prefetcher) state can optionally create an Update Object deriving it from PrefetcherStateUpdateType and pass the updates to the prefetcher.
    
    NextLinePrefetchEngine is an example implementation of PrefetchEngineIF
    DataPrefetcher is an example implementation of a simple prefetcher utilizing the above engine
    
    The files have been compile tested.
    Unit testing, LSU integration is TBD and will be done after review and possible changes.


core/DataPrefetcher.cpp
```diff
@@ -0,0 +1,121 @@
+#include "DataPrefetcher.hpp"
+#include "NextLinePrefetchEngine.hpp"
+
+namespace olympia
+{
+
+    DataPrefetcher::DataPrefetcher(sparta::TreeNode* node, const DataPrefetcherParameterSet* p) :
+        sparta::Unit(node),
+        PrefetcherIF<PrefetchEngineIF<>>(dynamic_cast<sparta::Unit*>(this)),
```
kathlenemagnus, line 9: Is there a way to avoid the dynamic cast? I think you can use the `getResourceAs` Sparta method like this:
```
node->getResourceAs<sparta::Unit*>
```

core/DataPrefetcher.cpp
```diff
@@ -0,0 +1,121 @@
...
+    {
+        req_queue_credits_out_.send(req_queue_.capacity());
+    }
+
+    //////////////////////////////////////////////////////////////////////
+    // callbacks
+
+    //! Process the incoming memory access and send to the engine
+    void
+    DataPrefetcher::processIncomingReq(const olympia::MemoryAccessInfoPtr & mem_access_info_ptr)
```
kathlenemagnus, line 49: Remove new line

jeffnye-gh (commented):
This code looks fine to me.

I've have a question on the API larger intent/future plans. I (will) put this in the originating issue [#142 ](https://github.com/riscv-software-src/riscv-perf-model/issues/142)

klingaard (commented):
Can you make a diagram on how these classes interrelate?  Give this tool a try: http://mermaid.js.org/

klingaard:
Hey @rajatbhatia1 where are you on this work?

core/DataPrefetcher.cpp
rajatbhatia1: I will remove the dynamic cast

core/DataPrefetcher.cpp
rajatbhatia1: sure

core/DataPrefetcher.cpp
```diff
@@ -0,0 +1,121 @@
...
+            // send access to output port and consume credit
+            sendPrefetch(access);
+            prefetcher_credits_--;
+
+            // schedule next prefetch generation (if any)
+            if (getPrefetchEngine()->isPrefetchReady() && (prefetcher_credits_ > 0))
+            {
+                ev_gen_prefetch_.schedule(sparta::Clock::Cycle(1));
+            }
+        }
```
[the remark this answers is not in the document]
rajatbhatia1: Actually, I just realize that I already checked for isPrefetchReady(), will remove this part of conditional. 
Here I am scheduling the next prefetch in queue (if I have credits)

klingaard:
I'm going to close this PR for now until @rajatbhatia1 has more updates for us to look at.

core/DataPrefetcher.cpp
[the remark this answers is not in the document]
rajatbhatia1: Scratch my earlier comment. This check is required as we have already handled one prefetch this cycle. However, there may or may not be pending prefetches already in queue which should be taken care of in subsequent cycles. 

There is no else, because this event will be scheduled in both cases of new prefetch requests being ready and prefetch credits coming back from the consuming unit

PR #181, opened 2024-07-17
Added Vector Uop Generation Support for Mask Generation and Widening Instructions

kathlenemagnus:
Continuing to add support for more and more vector instructions. 
- Expanded vector uop generation to support vector instructions with single destinations (e.g. integer compare instructions which generates masks) and vector widening instructions.
- Enabled several new categories of vector instructions by adding them to the vector uarch JSON
    - Vector Widening Integer Add/Subtract
    - Vector Bitwise Logical Instructions
    - Vector Integer Compare Instructions
    - Vector Integer Min/Max Instructions
    - Vector Widening Integer Multiply Instructions
    - Vector Mask-Register Logical Instructions
- Major refactor of the vector unit tester and improved the existing vector tests
- Added new tests for a widening vector instruction and a mask generation instruction
- Fixed bug in calculation to determine number of passes needed to execute a vector integer instruction (tail elements now considered)
- Added vector mask pipe target
- Added stat in the ROB for the total number of uops retired
- Refactored printing operators for pipe targets and stall reasons


kathlenemagnus:
@AaronGChan and @klingaard there is still a lot to do here, but I wanted to start this PR before it got any larger. I have a couple of questions for Aaron, but I'll put them in separate comments.

core/InstArchInfo.hpp
```diff
@@ -61,6 +62,22 @@ namespace olympia
...
 
+        using TargetPipeStringMap = std::map<TargetPipe, std::string>;
+        static const TargetPipeStringMap execution_pipe_string_map;
+
+        enum class UopGenType
+        {
+            ARITH,
+            ARITH_SINGLE_DEST,
+            ARITH_WIDE_DEST,
+            NONE
```
kathlenemagnus, line 73: Since NONE is a valid `UopGenType` (e.g. instructions like ignore LMUL like certain moves and mask compares), I need to add another type called UNKNOWN or INVALID.

core/ROB.cpp
```diff
@@ -159,6 +162,7 @@ namespace olympia
                 out_rob_retire_ack_rename_.send(ex_inst_ptr);
 
                 ++num_retired_;
+                num_uops_retired_ += ex_inst_ptr->getUOpCount();
```
kathlenemagnus, line 165: FYI I'm counting every instruction as 1 uop.

core/VectorUopGenerator.cpp
```diff
@@ -41,20 +94,48 @@ namespace olympia
     }
 
     const InstPtr VectorUopGenerator::generateUop()
+    {
+        const auto uop_gen_type = current_inst_->getUopGenType();
+        sparta_assert(uop_gen_type != InstArchInfo::UopGenType::NONE,
+            "Inst: " << current_inst_ << " uop gen type is unknown");
+        auto x = uop_gen_function_map_.at(uop_gen_type);
+        return x(this);
```
kathlenemagnus, line 102: Probably a cleaner way to do this. Open to suggestions...

test/core/dispatch/expected_output/big_core.out.EXPECTED
```diff
@@ -18,7 +18,7 @@
 {0000000000 00000000 top.dispatch info} Dispatch: mapping target: F2Iiq3
 {0000000000 00000000 top.dispatch info} Dispatch: mapping target: BRiq4
 {0000000000 00000000 top.dispatch info} Dispatch: mapping target: VINTiq5
-{0000000000 00000000 top.dispatch info} Dispatch: mapping target: VINTiq5
+{0000000000 00000000 top.dispatch info} Dispatch: mapping target: VSETiq5
```
kathlenemagnus, line 21: This change is just from a bug fix in the print operator.

test/core/vector/CMakeLists.txt
```diff
@@ -8,19 +8,24 @@ file(CREATE_LINK ${SIM_BASE}/arches     ${CMAKE_CURRENT_BINARY_DIR}/arches
...
-sparta_named_test(Vector_test_vsetvli_vl_max_setting Vector_test big_core.out -c test_cores/test_big_core_full.yaml --input-file vsetvli_vl_max_setting.json)
-sparta_named_test(Vector_undisturbed_test            Vector_test big_core.out -c test_cores/test_big_core_full_8_decode.yaml --input-file undisturbed_checking.json)
-sparta_named_test(Vector_unsupported_test            Vector_test big_core.out -c test_cores/test_big_core_full_8_decode.yaml --input-file vrgather.json)
+sparta_named_test(Vector_test_vsetivli      Vector_test big_core.out -c test_cores/test_big_core_full.yaml --input-file vsetivli_vaddvv.json)
+sparta_named_test(Vector_test_vsetvli       Vector_test big_core.out -c test_cores/test_big_core_full.yaml --input-file vsetvli_vaddvv.json)
+sparta_named_test(Vector_test_vsetvl        Vector_test big_core.out -c test_cores/test_big_core_full.yaml --input-file vsetvl_vaddvv.json)
+sparta_named_test(Vector_test_vsetivli_tail Vector_test big_core.out -c test_cores/test_big_core_full.yaml --input-file vsetivli_vaddvv_tail.json)
+#sparta_named_test(Vector_test_multiple_vset Vector_test big_core.out -c test_cores/test_big_core_full.yaml --input-file multiple_vset.json)
```
kathlenemagnus, line 27: @AaronGChan FYI this test wasn't being run before, but it was checked into the test repo so I enabled it. However, it currently fails in Rename so I need to look at it some more. Do you know why it's failing?

test/core/vector/Vector_test.cpp
```diff
@@ -37,84 +37,81 @@ const char USAGE[] = "Usage:\n"
 
 sparta::app::DefaultValues DEFAULTS;
 
-class olympia::DecodeTester {
+class olympia::DecodeTester
```
kathlenemagnus, line 40: @AaronGChan I'd really like your input on my changes to the vector tester. I made sure to keep the original intent of the tests in place, but I had to make a lot of changes to get them to pass.

test/core/vector/CMakeLists.txt
AaronGChan: Sorry, I think I forgot to uncomment it. I think it's failing due to timings, because I'm stopping the simulation at different time points to check the vset values. I guess a simpler solution would be to just split the test into multiple tests instead, that way we don't have to keep rerunning the simulation in the same test.

AaronGChan: I think we could just delete the test, and I could go in and add more tests related to it. I think with all the changes being added in recent PRs, the timings could have shifted by a cycle or two so debugging that older test probably isn't worth the effort.

test/core/vector/Vector_test.cpp
AaronGChan: I think your implementation is much cleaner! The json names would only need to be changed again if we add a lot more tests for varying vset values. In that case I could see it being necessary to include the vset values we're trying to set. 

Would it be useful at all to have a python script to autogenerate random vset tests to test edge cases? I think Knute and I had discussed this before to have a script for all vector instructions, so we could maybe expand the script as more functionality comes up. This is definitely more of a nice to have though and not pressing for getting vector support up.

test/core/vector/CMakeLists.txt
kathlenemagnus: I updated the test to include 2 vsets that use different LMUL values followed by 2 `vadd.vv` instructions that will each generate a different number of uops. Then the tester can check that the correct nnumber of uops were generated. Not a perfect test, but it'll do for now.

test/core/vector/Vector_test.cpp
kathlenemagnus: Good suggestion! I changed the names of the JSONs to include the vset config.

Yah I think auto-generating some tests would definitely be useful and would be easier to write in Python. Can you create an issue with your ideas?

core/InstArchInfo.hpp
kathlenemagnus: Fixed.

core/ExecutePipe.cpp
```diff
@@ -72,14 +75,14 @@ namespace olympia
             // will truncate, but we have each adder support the largest SEW possible
             if (ex_inst->getPipe() == InstArchInfo::TargetPipe::VINT)
```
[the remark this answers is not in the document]
AaronGChan: I think we can just support all under VALU adder, and then have a separate unit for masks, slides, and more complex computations like scatter/gather. I guess in the case of scatter/gathers, we just define a new unit that has a state machine to track for the VALU that is used only by that unit.

AaronGChan:
Did you want to remove the Uop checking code from rename in this PR, since you're already removing all old Uop requirements from ROB?
https://github.com/riscv-software-src/riscv-perf-model/blob/9e10a68f5b05abc485eb4ffadf938f7307437e94/core/Rename.cpp#L212

kathlenemagnus:
> Did you want to remove the Uop checking code from rename in this PR, since you're already removing all old Uop requirements from ROB?
> 
> https://github.com/riscv-software-src/riscv-perf-model/blob/9e10a68f5b05abc485eb4ffadf938f7307437e94/core/Rename.cpp#L212

Yes that would be great. We should just treat all instructions/uops the same way.

core/Inst.hpp
```diff
@@ -480,17 +465,16 @@ namespace olympia
         const bool is_csr_;
         const bool is_vector_;
         const bool is_return_;
-        bool has_tail_; // Does this vector instruction have a tail?
-        uint64_t uop_done_count_ =
-            1; // start at 1 because the uop count includes the parent instruction
-        uint64_t uop_count_ = 0;
+
         VCSRs VCSRs_;
+        bool has_tail_; // Does this vector uop have a tail?
```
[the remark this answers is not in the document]
kathlenemagnus: Fixed.

core/Inst.hpp
```diff
@@ -247,6 +247,14 @@ namespace olympia
 
         // Set the instruction's target PC (branch target or load/store target)
         void setTargetVAddr(sparta::memory::addr_t target_vaddr) { target_vaddr_ = target_vaddr; }
+        sparta::memory::addr_t getTargetVAddr() const { return target_vaddr_; }
+
+        void setVCSRs(const VCSRs * inputVCSRs)
+        {
+            VCSRs_.setVCSRs(inputVCSRs->vl, inputVCSRs->sew, inputVCSRs->lmul, inputVCSRs->vta);
```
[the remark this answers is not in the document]
kathlenemagnus: We would need to add a non-const get method, but yes we can do that.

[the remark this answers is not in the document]
kathlenemagnus: Fixed

kathlenemagnus:
@AaronGChan I'm going to go ahead and merge this PR. Can you remove the uop checking code from rename in your VLSU PR?

PR #183, opened 2024-07-18
VLSU Support

aarongchan:
PR Goal:
Implement VLSU instruction support. Current implementation uses the LSU design and adds vector iterations based on VLEN and data width of VLSU.


AaronGChan:
@kathlenemagnus 
If you could look at the [VLSU.cpp](https://github.com/riscv-software-src/riscv-perf-model/blob/47f1f861d26672a9e6a2500d44debb79279984bd/core/VLSU.cpp#L620) (specifically the `completeInst_`) for context for the rest of this comment when you get the chance. I've updated my code to work around how the LSU handled stores, but the issue I was running into and had to work around can be summarized as:
- Stores get retired before "completing" in the LSU. This is an issue because:
    - We need to send multiple memory requests through. If on vector store has 16 passes needed, so 16 memory operations, we need to send that many through. However, the wrappers for `LoadStoreInfo` and `MemoryAccessInfo`, which contain relevant data to cache, TLB, and MMU status include the `InstPtr`. This becomes an issue if we have multiple requests, because the `InstPtr` already has a retired status, but the next requests will try to set the status thus resulting in an error.
    - The work around I did was to create a separate status per each `LoadStoreInfo` wrapper, thus each memory request has it's own status, that is separate from the instruction's.
A couple questions:
- Currently in my design, the memory requests are sent through the VLSU serially, so we have to wait for the first pass to finish before the 2nd pass can begin for the same vector instruction. This is due to the current design of the LSU, because we're only assuming one memory request per instruction, which is the case for LSU, but not VLSU. The benefit of this is that if there are multiple vlsu instructions, they fully utilize the pipeline because while instruction A's memory request is waiting for the cache, instruction B's memory request is at a different stage of the vlsu pipeline. My question is does this type of design make sense? Or should rearchitect this to work differently similar to the uop generator in vlsu that Knute had suggested?
- The vlsu uop generator idea would only process one vlsu instruction at a time, but generate lets say 16 memory requests that then queue up in the VLSU. The flow would look like:
```
vlsu_uop_generator -> generates 16 memory requests for instruction A
vlsu_ldst_queue_ -> takes the memory requests, queues them up, begins processing them through the pipeline
vlsu -> once all memory requests are through, we officially retire instruction A
(repeat for all vlsu instructions)
vlsu_uop_generator -> generates 16 memory requests for instruction B
...

```

I wanted to check all this with you and Knute before proceeding, because it will change a bit of how the VLSU works compared to the LSU, especially around the setting of instruction status vs creating a `LoadStoreInfo` status due to the nature of VLSU instructions.

core/Decode.cpp
```diff
@@ -336,6 +336,9 @@ namespace olympia
         // uint32_t unfusedInstsSize = insts->size();
 
         // Decrement internal Uop Queue credits
+        ILOG(uop_queue_credits_)
```
kathlenemagnus, line 339: Did you mean to check these in?

core/Inst.hpp
```diff
@@ -485,6 +506,12 @@ namespace olympia
             1; // start at 1 because the uop count includes the parent instruction
         uint64_t uop_count_ = 0;
         VCSRs VCSRs_;
+        uint32_t eew_;
```
kathlenemagnus, line 509: I would move these values into the `VCSRs` structure and rename it to something like `VectorConfig` or `VectorState`.

core/Inst.hpp
```diff
@@ -485,6 +506,12 @@ namespace olympia
...
+        uint32_t mop_;
+        uint32_t stride_;
+
+        uint32_t vlsu_total_iters_ = 0;
```
kathlenemagnus, line 513: Also might want to consider an decorator like `RenameData` for `LSUData`.

core/Dispatch.cpp
```diff
@@ -237,7 +240,7 @@ namespace olympia
                                  "pipe. Did you define it in the yaml properly?");
             // so we have a map here that checks for which valid dispatchers for that
             // instruction target pipe map needs to be: "int": [exe0, exe1, exe2]
-            if (target_pipe != InstArchInfo::TargetPipe::LSU)
+            if (target_pipe != InstArchInfo::TargetPipe::LSU && target_pipe != InstArchInfo::TargetPipe::VLSU)
```
kathlenemagnus, line 243: Can you just check `isLoadStoreInst()` instead?

core/VectorUopGenerator.cpp
```diff
@@ -76,6 +76,14 @@ namespace olympia
         // Set weak pointer to parent vector instruction (first uop)
         sparta::SpartaWeakPointer<olympia::Inst> weak_ptr_inst = current_inst_;
         uop->setUOpParent(weak_ptr_inst);
+        uop->setEEW(current_inst_->getEEW());
+        uop->setMOP(current_inst_->getMOP());
+        uop->setStride(current_inst_->getStride());
+        if(uop->isLoadStoreInst()){
+            // set base address according to LMUL, i.e if we're on the 3rd
+            // LMUL Uop, it's base address should be base address + 3 * EEW
+            uop->setTargetVAddr(uop->getTargetVAddr() + uop->getEEW() * uop->getUOpID());
```
kathlenemagnus, line 85: This is fine for now since this PR only supports unit stride, but eventually we should create an "address unroller" class that generates all of the addresses for a vector uop. I think it makes sense to be a part of this class, but it could also be a part of the VLSU.

core/LoadStoreInstInfo.hpp
```diff
@@ -127,14 +134,44 @@ namespace olympia
...
+                // if UID is the same, check Uops for vector
+                return lhs->getInstUOpID() < rhs->getInstUOpID();
+            }
+            else
+            {
+                return lhs->getInstUniqueID() < rhs->getInstUniqueID();
+            }
+        }
+
+        void setVectorIter(uint32_t vec_iter){
```
kathlenemagnus, line 148: Open bracket should be on a new line.

core/LoadStoreInstInfo.hpp
```diff
@@ -127,14 +134,44 @@ namespace olympia
...
+        uint32_t getTotalVectorIter() const { return total_vector_iterations_; }
+
+        void setVLSUStatusState(Inst::Status vlsu_status_state ){ vlsu_status_state_ = vlsu_status_state; }
+        Inst::Status getVLSUStatusState(){ return vlsu_status_state_; } 
       private:
         MemoryAccessInfoPtr mem_access_info_ptr_;
         sparta::State<IssuePriority> rank_;
         sparta::State<IssueState> state_;
         bool in_ready_queue_;
+        uint32_t vector_iterations_ = 0;
```
kathlenemagnus, line 171: Is this really "completed vector iterations"? Could be renamed to be more clear.

core/LSU.cpp
```diff
@@ -258,17 +259,18 @@ namespace olympia
     {
         sparta_assert(inst_ptr->getStatus() == Inst::Status::RETIRED,
                       "Get ROB Ack, but the store inst hasn't retired yet!");
+        if(!inst_ptr->isVector()){
```
kathlenemagnus, line 262: You should assert if the inst is a vector since you have separate ports for scalar and vector.

core/LSU.cpp
```diff
@@ -1199,27 +1201,29 @@ namespace olympia
     // Update issue priority after store instruction retires
     void LSU::updateIssuePriorityAfterStoreInstRetire_(const InstPtr & inst_ptr)
     {
-        for (auto & inst_info_ptr : ldst_inst_queue_)
-        {
-            if (inst_info_ptr->getInstPtr() == inst_ptr)
+        if(!inst_ptr->isVector()){
```
kathlenemagnus, line 1204: Same comment as above. Just assert if vector since vector insts should never end up here.

core/MemoryAccessInfo.hpp
```diff
@@ -100,6 +100,13 @@ namespace olympia
             return inst_ptr == nullptr ? 0 : inst_ptr->getUniqueID();
         }
 
+        // This is a function which will be added in the SPARTA_ADDPAIRs API.
+        uint64_t getInstUOpID() const
```
kathlenemagnus, line 104: In what scenario is the inst ptr not set?

core/MemoryAccessInfo.hpp
```diff
@@ -151,6 +158,8 @@ namespace olympia
             replay_queue_iterator_ = iter;
         }
 
+        void setIsVector(bool is_vector){ is_vector_ = is_vector; }
+        bool isVector(){ return is_vector_; }
```
kathlenemagnus, line 162: Can we check the inst ptr instead?

core/ROB.cpp
```diff
@@ -73,19 +73,9 @@ namespace olympia
     {
         for (auto & i : *in_reorder_buffer_write_.pullData())
         {
-            if (!i->isUOp())
```
kathlenemagnus, line 76: I changed this code in my most recent PR so you may have conflicts whenever you sync your fork.

kathlenemagnus:
>The work around I did was to create a separate status per each LoadStoreInfo wrapper, thus each memory request has it's own status, that is separate from the instruction's.

I think this is the right solution. Each memory access should be completed separately and then a vector instruction can be marked completed if all of its memory accesses have completed.

>Currently in my design, the memory requests are sent through the VLSU serially, so we have to wait for the first pass to finish before the 2nd pass can begin for the same vector instruction. This is due to the current design of the LSU, because we're only assuming one memory request per instruction, which is the case for LSU, but not VLSU. The benefit of this is that if there are multiple vlsu instructions, they fully utilize the pipeline because while instruction A's memory request is waiting for the cache, instruction B's memory request is at a different stage of the vlsu pipeline. My question is does this type of design make sense? Or should rearchitect this to work differently similar to the uop generator in vlsu that Knute had suggested?

I would rearchitect like Knute suggested. What we want is for a vector instruction to be able to generate multiple memory accesses. They can be executed serially, but they should be able to be pipelined. So 1 vector instruction should be able to send a uop down the LSU pipeline every cycle. Similarly to how we used to track whether a parent instruction was "done" in the ROB, a vector load or store should expect multiple memory accesses to be completed before it can be marked as done. One thing to be careful of though is that there should only be 1 writeback to the vector destination. So you will need some sort of structure for collecting all of the data returned to the LSU.

>The vlsu uop generator idea would only process one vlsu instruction at a time, but generate lets say 16 memory requests that then queue up in the VLSU. The flow would look like:
>
>>vlsu_uop_generator -> generates 16 memory requests for instruction A
vlsu_ldst_queue_ -> takes the memory requests, queues them up, begins processing them through the pipeline
vlsu -> once all memory requests are through, we officially retire instruction A
(repeat for all vlsu instructions)
vlsu_uop_generator -> generates 16 memory requests for instruction B
>
>I wanted to check all this with you and Knute before proceeding, because it will change a bit of how the VLSU works compared to the LSU, especially around the setting of instruction status vs creating a LoadStoreInfo status due to the nature of VLSU instructions.

Let's talk about this more on Monday. I see two paths forward here:
1. We could do as you suggest and "sequence" the vector load store instruction when it gets to the LSU and generate a memory request that is stored in the LDST queue. These memory requests would behave like scalar instructions in the queue. The question here is when to trigger WB. It's inefficient to writeback to the vector destination 16 times and it is not always possible to do partial writes to the register file.
2. Another option would be to keep the vector instruction in the LDST queue and send it down the LSU pipe multiple times. Each time it gets sent down the pipe it would generate a different memory access. Again, we need a way to track all of the data that has come back and make sure it's all ready before sending the vector instruction down the pipe a final time to do writeback.

aarongchan:
@klingaard @kathlenemagnus this should be ready for review/merge.

klingaard (approved):
I'll preemptively approve this as Kathlene will be leading the charge on this PR for now. 

Thank you for your contributions to the project, @aarongchan!  Best of luck to you at Intel.

aarongchan:
Closing due to @kathlenemagnus's [PR](https://github.com/riscv-software-src/riscv-perf-model/pull/219)

PR #184, opened 2024-07-18
Vector Uop Generator Bug

AaronGChan:
The following [code](https://github.com/AaronGChan/riscv-perf-model/blob/5f86ada39301c3181e47236a3c592a5961a48e99/core/Decode.cpp#L274):
```vec_uop_gen_->getNumUopsRemaining() > 1```
doesn't perform correctly for LMUL=2, because the `num_uops_to_generate_` will be decremented to 1 on the `setInst` of `VectorUopGenerator`, so the 2nd uop is never generated.

This PR fixes logic and counting of Uops based on the parent instruction.

### DIFF
--- a/core/Decode.cpp
+++ b/core/Decode.cpp
@@ -271,7 +271,7 @@ namespace olympia
                     // Original instruction will act as the first UOp
                     inst->setUOpID(0); // set UOpID()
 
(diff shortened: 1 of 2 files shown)


klingaard (approved):

kathlenemagnus (approved):
I think I fixed this in my other PR but I'll confirm that LMUL 2 works before I merge.

PR #188, opened 2024-07-23
Update to latest stf_lib

### LINKED ISSUE #185: Build problems
Hi, I am starting on Olympia and having some problems with respect the build of various components which I already solved.

The last thing that I am doing is trying to generate a trace via Dromajo with drystone benchmark. But it seems that it needs a backward version of gcc/glibc.

Do you have a version which for the moment I can build Dromajo?

Thanks!

bdutro:
Fixes #185


klingaard (approved):

bdutro:
@klingaard you'll need to hit the merge button since I don't have write permission.

PR #190, opened 2024-07-31
Added More Vector Arithmetic Instruction Support

kathlenemagnus:
Added support for even more vector arithmetic instructions:
- Vector Single-Width Shift Instructions
- Vector Narrowing Integer Right Shift Instructions
- Vector Single-Width Integer Multiply-Add Instructions
- Vector Widening Integer Multiply-Add Instructions
- Vector Integer Merge Instructions
- Vector Integer Move Instructions

Also redesigned how the vector uop generator handles vector instructions when only 1 uop is needed (e.g, LMUL = 1, certain vector instructions are not sequenced). The vector uop generator will always generate at least one uop which will be used in place of the "parent" instruction. This is necessary because some vector instructions require the source and/or dest registers be added. For example, the multiply-add vector instructions need to have their destination added as one of their sources.

There are two categories of vector arithmetic instructions that cannot be supported correctly because of issues with Mavis:
- Vector Integer Extension
- Vector Integer Add-with-Carry/Subtract-with-Borrow Instructions

Once these Mavis issues are resolved, all vector arithmetic instructions will be supported by Olympia.

### DIFF
--- a/core/Inst.hpp
+++ b/core/Inst.hpp
@@ -215,11 +215,6 @@ namespace olympia
         // UID, but different UOp IDs.
         uint64_t getUOpID() const { return uopid_.isValid() ? uopid_.getValue() : 0; }
 
-        bool hasUOps() const { return uopid_.isValid() && uopid_.getValue() == 0; }
-
-        // UOpIDs start at 1, because we use 0 as default UOpID on initialization
-        bool isUOp() const { return uopid_.isValid() && uopid_ > 0; }
-
         void setBlockingVSET(bool is_blocking_vset) { is_blocking_vset_ = is_blocking_vset; }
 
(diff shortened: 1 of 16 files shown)


core/Inst.hpp
```diff
@@ -341,6 +334,27 @@ namespace olympia
                 });
         }
 
+        uint64_t hasImmediate() const { return opcode_info_->hasImmediate(); }
+        uint64_t getImmediate() const { return opcode_info_->getImmediate(); }
+
+        bool getVectorMaskEnabled() const
```
klingaard, line 340: Can this method be cached in the base Inst?

klingaard (approved):

PR #192, opened 2024-08-05
Using Vectors to support multiple pipelines

Forchapeatl:
-  Introduced support for multiple load/store pipelines by using vectors for queues, pipelines, and ready queues. Each pipeline can be managed independently.
  
- Added mutexes (`cache_mutex_` and `tlb_mutex_`) to manage concurrent access to shared resources like cache and MMU/TLB.

- Updated methods to handle operations per pipeline, like `allocateInstToIssueQueue_`, `appendToReplayQueue_`, and `popIssueQueue_`, which now include `pipeline_id` to specify which pipeline is being accessed.

### DIFF
--- a/core/LSU.hpp
+++ b/core/LSU.hpp
@@ -1,4 +1,3 @@
-
 #pragma once
 
 #include "sparta/ports/PortSet.hpp"
@@ -31,6 +30,9 @@
 #include "MMU.hpp"
 #include "DCache.hpp"
 
+#include <vector>
+#include <mutex>
+
 namespace olympia
 {
     class LSU : public sparta::Unit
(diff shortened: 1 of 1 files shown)


klingaard:
This is awesome, but I feel there are files missing in this PR.  Shouldn't the source files be included?  Maybe I'm missing something...

core/LSU.hpp
klingaard, line 1: Are the mutexes necessary?  Did you add threading support to LS (which would be cool)?  If you're concerned about asynchronous behaviors between the pipes, the Sparta Framework isn't threaded, so it will most likely call each pipeline in a round-robin fashion.

klingaard:
@Forchapeatl can you provide the missing files for this PR?

klingaard:
Closing this PR.  Thank you for your submission!

PR #198, opened 2024-08-12
Add support to multiple load store pipelines

orion160:
TODO:
- add tests

### DIFF
--- a/core/LSU.cpp
+++ b/core/LSU.cpp
@@ -297,7 +297,8 @@ namespace olympia
         ++lsu_insts_issued_;
 
         // Append load/store pipe
-        ldst_pipelines_[0]->append({win_ptr, 0});
+        ldst_pipelines_[ldst_active_pipeline_idx]->append({win_ptr, ldst_active_pipeline_idx});
+        ldst_active_pipeline_idx = (ldst_active_pipeline_idx + 1) % ldst_pipeline_count_;
 
         // We append to replay queue to prevent ref count of the shared pointer to drop before
         // calling pop below
@@ -324,14 +325,16 @@ namespace olympia
 
     void LSU::handleAddressCalculation_(const LSPayload & ls_payload)
     {
+        auto* ldst_pipeline = ldst_pipelines_[ls_payload.ldst_pipeline_idx].get();
+
         auto stage_id = address_calculation_stage_;
 
-        if (!ldst_pipelines_[0]->isValid(stage_id))
+        if (!ldst_pipeline->isValid(stage_id))
         {
             return;
         }
 
-        auto & ldst_info_ptr = (*ldst_pipelines_[0])[stage_id].ldst_inst_info_ptr;
+        auto & ldst_info_ptr = ls_payload.ldst_inst_info_ptr;
         auto & inst_ptr = ldst_info_ptr->getInstPtr();
         // Assume Calculate Address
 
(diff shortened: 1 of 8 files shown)


core/LSU.hpp
```diff
@@ -164,8 +167,23 @@ namespace olympia
...
         // Load/Store Pipeline
-        using LoadStorePipeline = sparta::Pipeline<LoadStoreInstInfoPtr>;
-        LoadStorePipeline ldst_pipeline_;
+        const uint32_t ldst_pipeline_count_;
+        uint32_t ldst_active_pipeline_idx = 0;
+
+        struct LSPayload
+        {
+            LoadStoreInstInfoPtr ldst_inst_info_ptr;
+            uint32_t ldst_pipeline_idx;
```
klingaard, line 176: Pls initialize or make `const`:
```suggestion
            uint32_t ldst_pipeline_idx = std::numeric_limits<uint32_t>::max();
```

core/LSU.hpp
```diff
@@ -164,8 +167,23 @@ namespace olympia
...
+        };
+
+        friend inline std::ostream & operator<<(std::ostream & os, const LSPayload & ls_payload)
+        {
+            return os;
```
klingaard, line 181: You're 90% there -- might as well do this:
```suggestion
            os << ls_payload.ldst_inst_info_ptr << std::endl;
            return os;
```

core/LSU.cpp
```diff
@@ -364,7 +382,15 @@ namespace olympia
             }
         }
 
-        out_mmu_lookup_req_.send(mem_access_info_ptr);
+        // TODO: set a max upper bound
+        uint32_t mmu_sync_delay = 0;
+        while (out_mmu_lookup_req_.isDriven(mmu_sync_delay))
+        {
+            mmu_sync_delay++;
+        }
```
klingaard, line 390: I kinda know what you're doing here -- and I kinda don't.  :laughing:   Are you looking to see how many outstanding requests there are?

core/LSU.cpp
```diff
@@ -373,7 +399,9 @@ namespace olympia
         const auto stage_id = mmu_lookup_stage_;
 
         // Check if flushing event occurred just now
-        if (!ldst_pipeline_.isValid(stage_id))
+
+        // TODO - attach pipeline idx in response
+        if (!ldst_pipelines_[0]->isValid(stage_id))
```
klingaard, line 404: Is this unfinished code?

core/LSU.cpp
```diff
@@ -515,7 +545,14 @@ namespace olympia
             return;
         }
 
-        out_cache_lookup_req_.send(mem_access_info_ptr);
+        // TODO: set a max upper bound
+        uint32_t cache_sync_delay = 0;
+        while (out_cache_lookup_req_.isDriven(cache_sync_delay))
```
klingaard, line 550: Same question as above for the MMU.  Are you looking to stagger the requests?

core/LSU.cpp
```diff
@@ -750,8 +791,12 @@ namespace olympia
         lsu_flushes_++;
 
         // Flush load/store pipeline entry
-        flushLSPipeline_(criteria);
 
+        // TODO: add logic for single pipeline flush?
```
klingaard, line 795: Comment is questionable -- if you have a transaction in flight that needs to be flushed, it technically can be in ANY pipe.

core/LSU.cpp
orion160: My thoughts on the synchronization problem was that multiple ldst pipelines could send requests at the same time

klingaard: Ok, so you're saying they can _make requests_ at the same time, but the requests are staggered coming into the MMU.  

I think this will break the LSU pipeline concept.  Specifically, if a load is traversing the LS pipeline and the pipeline is "programmed" to stage it in 4 cycles: 

1. agen
2. xlate
3. cache
4. drive

You're artificially delaying stage 2 with a late request to the MMU, so when the load hits the cache stage of the pipeline, it will see the PA is not ready and drop the load: https://github.com/riscv-software-src/riscv-perf-model/blob/33d2e2fbc99a3439b5d15ee9220558ca1467d909/core/LSU.cpp#L437

PR #210, opened 2024-09-26
Added Conda Environment Yaml for MacOS

kathlenemagnus:
I have never been able to successfully build the Olympia Conda environment on my Mac, so I'm checking in a yaml that works for me. It's derived from the Sparta Conda environment with some modifications (i.e. removing version specifications from a few libraries like boost).


jeffnye-gh:
Sorry I can't try this out, I don't use macs.

jeffnye-gh (approved):
I'm approving so I don't block the process, but I do not have a mac, i can not try this.

kathlenemagnus:
I will wait to merge until another MacOS user can test this out.

klingaard:
Ugh, don't be too frustrated with this, but we actually have a MacOS environment yaml file here: https://github.com/riscv-software-src/riscv-perf-model/blob/master/.github/workflows/conda/macos_env.yml

Moving forward, can you either merge your new file into it OR have the GitHub CI use yours?  This is the line in CI that you'd update: https://github.com/riscv-software-src/riscv-perf-model/blob/791d133283861273f3649c0e05aedccd946b29a3/.github/workflows/macos-build.yml#L72

PR #212, opened 2024-10-08
Various bug fixes and cleanup

klingaard:
- DIspatch stall counts were off by one stall condition providing incorrect stall counts
- The handshake between the LS and Data was off by one cycle forcing the LS to always see a cache miss even when hit
- Some basic cleanup on the code


klingaard:
The bug fix on dcache response.  This is dhrystone with always hit DTLB and Dcache.

Before:
image


After:
image

kathlenemagnus (approved):

lumicrespo:
It seems there is an issue with the layout files:

1. The gen_alf.py script fails to correctly identify or find the execution units (alu, fpu, etc), which results in an incomplete layout generation.
1. The Makefile appears to only account for the pipeout of the small_arch configuration, potentially ignoring other architectural setups.

PR #228, opened 2024-11-19
[Vector] refactor vuop

orion160:


kathlenemagnus (approved):
This is great work, @orion160!

PR #233, opened 2024-12-02
Design document of the Branch Prediction Unit

dragon540:
This is the design document for the branch prediction unit to be added to Olympia simulator. This document aims to give an overview of the micro-architectural and implementation detail of the BPU. 

Document will be further updated as development progresses. Any suggestions are appreciated.

### DIFF
--- a/docs/bpu_design_document/BPUDesignDoc.adoc
+++ b/docs/bpu_design_document/BPUDesignDoc.adoc
@@ -53,66 +53,27 @@ which are used to update the BPU once instructions are finally committed.
 [[OVERVIEW]]
 == OVERVIEW
 
-Branch Prediction Unit is the unit responsible for predicting the branch
-outcomes and branch targets before branch instructions are actually
-resolved in the pipeline. This prediction allows the processor to prefetch and execute subsequent instructions without waiting to resolve the branch condition, minimizing stalls and maintaining high throughput in the pipeline.
+Branch Prediction Unit is the unit responsible for predicting the branch outcomes and branch targets before branch
+instructions are actually resolved in the pipeline. This prediction allows the processor to prefetch and execute
+subsequent instructions without waiting to resolve the branch condition, minimizing stalls and maintaining high
+throughput in the pipeline.
 
-It receives prediction input in the form of PC and instruction type from the
-Fetch unit and predicts the direction and target address from which Fetch unit
-has to get the next instruction.
+The proposed design receives input in the form of `PredictionRequest` from the Fetch unit and predicts in the form
+of `PredictionRequest` (includes direction and target address) from which Fetch unit fetches the next instruction.
+Information to update the internal state of BPU is received in the form of `UpdateInput` when the instructions are
+committed.
 
-[[Functional_Description]]
-== Functional Description
+Olympia's Branch Prediction API expects the implementation of Branch Predictor to specify its implementation of a
+prediction input, prediction output, and the update input.
 
-The proposed Branch Prediction Unit (BPU) is a two-tiered branch predictor where the
-first tier(BasePredictor) provides a simple but fast prediction, whereas the second tier(TAGE_SC_L)
-can give more accurate prediction but takes an additional cycle.
-
-[[Counters]]
-=== Counters
-
-** `pred_req_num` - Total number of prediction requests made to BPU
-** `num_mispred` - Total number of mis-predictions
-** `mispred_perc` - Percentage of mis-predictions
-** `branch_req_num` - Total number of prediction requests made which are conditional branch instructions.
-** `call_req_num` - Total number of prediction requests made which are call/jump type.
-** `return_req_num` - Total number of prediction requests made which are return type.
-** `pht_req_num` - Total number of requests made to PHT.
-** `pht_hit_num` - Total number of times when requested index is hit on PHT.
-** `pht_miss_rate` - Total number of times when requested index is miss on PHT.
-** `pht_mispred_perc` - Percentage of times when PHT mispredicted. 
-** `btb_req_num` - Total number of requests made to BTB.
-** `btb_hit_rate` - Total number of times when requested index is hit on BTB.
-** `btb_miss_rate` - Total number of times when requested index is miss on BTB.
-** `ras_high_mark` - Highest size of RAS during an execution cycle.
-** `ras_low_mark` - Lowest size of RAS during an execution cycle.
-
-[[Unit_Block_Diagram]]
-=== Unit Block Diagram
-
-image:media/bpu_uarch.png[image,width=800,height=366]
-
-Figure 1 - Unit block diagram of BPU
-
-[[Block_Diagram_Description]]
-=== Block Diagram Description
-
-BPU consists of a BasePredictor, a TAGE_SC_L-Predictor, and an FTQ (a
-queue) which stores the information of all in-flight branch instructions
-and is required in order to update the state of BPU once instruction are finally
-committed, it also enables multi-tier prediction by buffering the initial prediction
-while BPU accesses its higher level of hierarchy.
-
-[[Description_of_PredictionRequest]]
-== Description of PredictionRequest
-
-Olympia's Branch Prediction API expects the implementation of Branch Predictor to specify its implementation of a prediction input.
+[[PredictionRequest]]
+== PredictionRequest
 
 [[Overview_of_PredictionRequest]]
 ===  Overview
-PredictionRequest is received by the BPU from the Fetch unit whenever a request for
-the prediction is made. Based on the data provided by this input, BPU makes the
-prediction.
+
+`PredictionRequest` is received by the BPU from the Fetch unit whenever a request for the prediction is made. Based on
+the data provided by this input, BPU makes the prediction.
 
 [[Structure_of_PredictionRequest]]
 ===  Structure of PredictionRequest
@@ -123,7 +84,9 @@ prediction.
 [[Class_Description]]
 ===  Class Description
 
-==== Constructor -
+`class PredictionRequest`
+
+==== Constructor
 `PredictionRequest(uin64_t PC, uint8_t instType) : PC_(PC), instType_(instType) {}`
 
 === Data members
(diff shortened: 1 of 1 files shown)


docs/bpu_design_document/BPUDesignDoc.adoc
```diff
@@ -0,0 +1,636 @@
...
+committed.
+
+[[Description_of_PredictionInput]]
+== Description of PredictionInput
+
+Olympia's Branch Predictor API intends the implementation of Branch Predictor to
+define custom PredictionInput
+
+[[Overview_of_PredictionInput]]
+===  Overview
```
jeffnye-gh, line 134: Are there header formatting issues with the document? The organization from this point on does not follow a progression I understand.

jeffnye-gh (changes requested):
There are organization changes I would suggest. I would compare this to the template. I would like a better understanding of your class sets and their hierarchy, their methods.

I would like to get a better understanding of how you intend the unit to operate and a sense of how much code needs to be written. With some mention of the existing classes and APIs.

Global comments;
take advantage of the adoc syntax when inserting code fragments to help readability.
Be sure to document your material references when external information is used. Such as figures etc.

jeffnye-gh:
This is similar to Shobhit's other PR. Not expected to compile.

We asked that interns post their code in what ever state as draft PRs at the end of the internship so as not to lose any progress.

dragon540:
@klingaard @arupc this document is still a work in progress however some details are much more clearer than before. Can you take a look at this, and check if the high level design of the unit (interaction of BPU with Fetch and FTQ, organization of its constituent predictors, etc) looks okay?

klingaard:
> @klingaard @arupc this document is still a work in progress however some details are much more clearer than before. Can you take a look at this, and check if the high level design of the unit (interaction of BPU with Fetch and FTQ, organization of its constituent predictors, etc) looks okay?

Can do!

oms-vmicro:
@dragon540 I know some of the code development is already being done in #243 but providing some feedback specific to the design document.

Overall I think the document is well done and covers a lot of the components one would like to see in a more high performance BPU & Front-end.

At a high level, my understanding from the design doc is the `BranchPredIF.hpp` interface is intended to be used for the BPU unit.  `PredictionRequest`, `PredictionOutput`, and `UpdateInput` correspond to the interface "prediction input", "prediction output" and "update input" types from the interface.

I don't believe the `BranchPredIF.hpp` interface is currently being used by the individual branch predictors themselves though (e.g. PHT, BTB, TAGE, etc).  I believe the intent behind the interface is that individual predictors can and should use the full interface as well.  Specifically, inputs and outputs to the individual predictors should be templatized such that they can be redefined as needed.  The existing predictors demonstrate where this would be useful as the PHT takes a hash of PC and global history as input vs PC being used for the other predictors.  With using predictor-specific `PredictionRequest`, `PredictionOutput`, and `UpdateInput` types, a generic interface can be used with any individual predictor and defined to suit that predictor (e.g. PHT -> hash of PC & global hist, BTB -> PC).  They can then be defaulted as is currently defined by the design doc.  This would make for a less "hardened" approach to interfacing with the individual predictors along with the interfaces to the overall BPU unit.

In a similar thought, if there is a way to enable or disable individual predictors via command-line or configuration, I think that would be helpful.  Meaning, an ability to define the collection of actual predictors implemented in the BPU would be useful along with a modular branch-predictor approach so then using a configuration file one can specify the collection of predictors implemented in the BPU (e.g. a BOOM uarch configuration, etc).  The thought being that one can then enable, disable, or even add additional modular predictors that can be selected at via configuration.  A simple example would be with the existing design, if one wanted to evaluate TAGE with and without the Statistical Corrector predictor.  This would likely mean having classes for the individual predictors and then parent classes coupling them into "BasePredictor" and "TAGE_SC_L".

- I realize that brings up a question of how does the design function/what assumptions can be made if someone were to disable all individual predictors, but I think we can prevent that and assume there will be one "base" (simple) and possibly one "TAGE_SC_L" (complex) predictor.

docs/bpu_design_document/BPUDesignDoc.adoc
```diff
@@ -0,0 +1,1058 @@
...
+3. `bool isHit(uint64_t PC)` - returns whether there is an entry for the PC in BTB or not.
+
+4. `uint64_t getPredictedPC(uint64_t PC)` - returns the target PC corresponding to the
+PC
+
+==== Private functions
+None
+
+[[Return_Address_Stack]]
+== Return Address Stack
```
oms-vmicro, line 486: I know BOOM uses a similar name for their return address predictor, but we might want to consider avoiding the acronym `RAS`.

When I see the term `RAS` with respect to computing I think of features related to "Reliability, Availability, and Servicability" ([wiki](https://en.wikipedia.org/wiki/Reliability,_availability_and_serviceability)).  You'll find the acronym used in this manner in a variety of places: 

- https://github.com/riscv-admin/riscv-ras/blob/main/Specification.adoc
- https://www.kernel.org/doc/html/v4.10/admin-guide/ras.html
- https://developer.arm.com/documentation/107790/0100/Introduction-to-RAS
- https://www.intel.com/content/www/us/en/developer/articles/technical/pmem-RAS.html

Not dead set on this, but to potentially avoid confusion we could consider naming it something different.  Some suggestions:

- Return Stack Buffer (`RSB`): https://en.wikipedia.org/wiki/Branch_predictor#Prediction_of_function_returns
- Return Address Predictor (`RAP`)
- Return Address Buffer (`RAB`)

docs/bpu_design_document/BPUDesignDoc.adoc
```diff
@@ -0,0 +1,1058 @@
...
+
+=== Overview
+This section describes the overview and description for the top level class
+which bolts the Branch Prediction Unit into the Olympia simulator.
+
+[[Functional_Description]]
+=== Functional Description
+
+The proposed Branch Prediction Unit (BPU) is a two-tiered branch predictor where the
+first tier(BasePredictor) provides a simple but fast prediction, whereas the second tier(TAGE_SC_L)
```
oms-vmicro, line 173: This is a suggestion, but wanted to see if we would want to use terms other than `first` and `second` to describe the branch predictor tiers here and in the interfaces/functions below.

The thought was if there is a more defined way of describing these tiers that is still generic but provides some guidance as to what predictors fit in which tier.  My suggestion would be to potentially call them "fast" vs "slow" or "single-cycle" vs "multi-cycle" as I believe first-tier predictors are expected to complete in a single cycle whereas second-tier predictors take more than one cycle.

docs/bpu_design_document/BPUDesignDoc.adoc
```diff
@@ -0,0 +1,1058 @@
...
+sent out. Otherwise, an entry is made to the BTB.
+
+** If the instruction type is a "ret", then the PC is simply popped from
+the RAS, and the PC is sent out.
+
+** If the instruction type is a branch then the PC is sent simultaneously
+to the BTB, BasePredictor and the TAGE_SC_L-Predictor.
+*** If it is a hit on BTB, and the BasePredictor predicts a taken
+branch, then the output is sent to Fetch unit
+*** If it is a hit on BTB, but
```
oms-vmicro, line 345: Complete sentence.  Possibly: "... but the BasePredictor predicts a not taken branch, then fall-through is sent to Fetch unit".

docs/bpu_design_document/BPUDesignDoc.adoc
```diff
@@ -0,0 +1,1058 @@
...
+
+==== Public data members
+None
+
+==== Private data members
+
+1. `const uint32_t pht_size_`
+2. `const uint8_t  ctr_bits_`
+3. `const uint8_t  ctr_bits_val_`
+4. `std::map<uint64_t, uint8_t> pht_`
```
oms-vmicro, line 425: I believe PHT will use a `uint32_t` index representing a hash of PC and global history.  I think this is noted by the use of `uint32_t` in the increment, decrement, and getPrediction functions noted below.  

Should the key/index for the PHT `std::map` here also be a `uint32_t`?  (e.g. `std::map<uint32_t, uint8_t> pht_`)

docs/bpu_design_document/BPUDesignDoc.adoc
```diff
@@ -0,0 +1,1058 @@
...
+tage_base_ctr_bits_(tage_base_ctr_bits) {}`
+
+* Initializes counter at all indexes of BIM with 0
+
+=== Data members
+
+==== Private data members
+1. `uint32_t tage_bim_max_size_` - Represents maximum size of the BIM table of TAGE
+2. `uint8_t tage_bim_ctr_bits_` - Represents the number of bits used in counter of BIM table
+3. `std::vector<uint8_t> Tage_Bimodal_` - Represents the container used for BIM in the
```
oms-vmicro, line 705: Should this be `uint32_t` if `tage_bim_table_size_` is a `uint32_t` as defined above?

jeffnye-gh:
From todays call 2025.06.19, regarding changes to the api to support staged predictions. 

Short of it is: I think we want to expose signals in the API to allow override of previous predictions. This has some implications to what is behind the BP-API.

In recent open source designs there is the concept of staged prediction, the uBTB/LP/TAGE/SC/ITTAGE all deliver predictions, some with longer latency. Not all of these will exist, not all latency values will be different. 

In terms of the visible methods in the BP-API I think you want to add a method to signal an override to the front end logic and pass a struct with override info, BP request queue idx, basic block address, new prediction, maybe an extension for debug/metadata info.

I am making an assumption that the decision to signal an override has been made behind the BP-API, a topic for discussion, but seems like a way to allow greatest freedom/generality to developers to explore choices and keep the API smaller.

You could imagine I would test the effectiveness of a tournament selector vs simply always choosing the latest prediction that is different than an earlier one.

One other thing I did not mention in the call, I believe the BP-API will want to support multiple prediction requests, updates, and returns. If you read the trade press 2 or 4 predictions at a time will become the bar. I believe this just means that request/result data is grouped. Not a significant change on the surface, but occurred to me while writing.

klingaard:
@dragon540 there is a lot of great feedback from @jeffnye-gh and @oms-vmicro in this PR.  Can you resolve the conversations that you have addressed and comment/fix the remaining issues?  When you're ready, please convert this draft PR and we can get it merged in.  Thanks

PR #236, opened 2024-12-05
unit test template

jeffnye-gh:
A unit test bench template. It intends to help new users create unit tests.

Three components, Src, Sink, Dut. Handles json and stf input test types.

Conforms to coding style through clang-formt.

I created this for myself a while back to help learning the concepts, updated for coding style.

It was useful to me when learning sparta/map/olympia.


jeffnye-gh:
this fails for MAC OS but I do not see a connection between my PR and the mac failure. I believe it fails due to the warning installing conda. 

Warning: Future Miniforge releases will NOT build Mambaforge installers. We advise you switch to Miniforge at your earliest convenience. More details at [https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/.](https://conda-forge.org/news/2024/07/29/sunsetting-mambaforge/)
  Warning: ERROR: executing pre_install.sh failed

jeffnye-gh:
re-running the checks with change, it now works.

klingaard (approved):
This is super awesome.  What are your thoughts about moving this into Sparta?

jeffnye-gh:
Hi Knute.

Yes, moving it to sparta is fine with me.  Couple questions.

Does this collide with SkeletonPipeline? Honestly I'm only now noticing
that sparta example and the similarities, my bad.

If moved into sparta should it be only there, or in both olympia and sparta
?

Your call.

Jeff

klingaard:
I don't think it collides with anything in Sparta, really.  If we do put this in sparta, a set of cool scripts would be nice to set up a unit.  But I think here is fine.

PR #238, opened 2024-12-08
[vuop_gen]

orion160:
Also implements integer extension instructions which requires the value encoded in the instruction for sequencing.


test/core/vector/CMakeLists.txt
```diff
@@ -29,3 +29,46 @@ sparta_named_test(Vector_test_vmulvv        Vector_test -l top info vmulvv.out
 sparta_named_test(Vector_test_vmseqvv       Vector_test -l top info vmseqvv.out       -c test_cores/test_big_core.yaml --input-file vmseqvv_e8m4.json)
 sparta_named_test(Vector_test_vsadd         Vector_test -l top info vsadd.out         -c test_cores/test_big_core.yaml --input-file vsadd.json)
 sparta_named_test(Vector_unsupported_test   Vector_test -l top info unsupported.out   -c test_cores/test_big_core.yaml --input-file vrgather.json)
+
+file(CREATE_LINK ${CMAKE_CURRENT_SOURCE_DIR}/elementwise.json ${CMAKE_CURRENT_BINARY_DIR}/elementwise.json SYMBOLIC)
```
kathlenemagnus, line 33: I'm confused about the changes to the tester. Some of these tests seems like they're just renamed versions of existing tests and other tests seem new. And we really only need 1 test binary. Can you merge `Vector_test.cpp` and `Vuop_test.cpp`? We only need to check the stats from the `VectorUopGenerator` and `Retire`.

arches/isa_json/gen_uarch_rv64v_json.py
```diff
@@ -188,7 +186,6 @@
      "vwmaccus.vx" : {"pipe" : "vmul", "uop_gen" : "MAC_WIDE", "latency" : 3},
 
 # Vector Integer Arithmetic Instructions: Vector Integer Merge Instructions
-# FIXME: Requires Mavis fix to include vector mask
```
kathlenemagnus, line 191: This hasn't been fixed - please add the comment back.

core/VectorUopGenerator.cpp
```diff
@@ -117,8 +150,10 @@ namespace olympia
...
         num_uops_to_generate_ = std::ceil(vector_config->getVLMAX() / num_elems_per_uop);
 
-        if ((uop_gen_type == InstArchInfo::UopGenType::WIDENING)
-            || (uop_gen_type == InstArchInfo::UopGenType::MAC_WIDE))
+        if (uop_gen_type == InstArchInfo::UopGenType::WIDENING
+            || uop_gen_type == InstArchInfo::UopGenType::WIDENING_MIXED
+            || uop_gen_type == InstArchInfo::UopGenType::MAC_WIDE
+            || uop_gen_type == InstArchInfo::UopGenType::NARROWING)
         {
             // TODO: Add parameter to support dual dests
```
kathlenemagnus, line 158: It would be a good idea to add a `sparta_assert` to make sure that LMUL <= 4 for these categories.

kathlenemagnus (approved):
@orion160 this is excellent! I just have a few suggestions regarding the `Modifier` class you added to support the int extension instructions and the tester. Great job!

test/core/vector/CMakeLists.txt
orion160: Yes, I was thinking on the Vuop_test to test the different instruction modes meanwhile the Vector_test does vector in a more general way.

I'll fuse them and make more exhaustive test with more configurations (different LMUL and SEW values)

PR #240, opened 2024-12-16
Branch prediction unit implementation

dragon540:
This PR is intended for the implementation of the BPU as per the design document (see PR #233). 

Only intended to save test bench, unit structure, etc to get back to implementation work after a short while, and is not building currently.


jeffnye-gh:
This is a draft PR it is not expected to compile, Shobhit mentions this.

"Only intended to save test bench, unit structure, etc to get back to implementation work after a short while, and is not building currently."


We asked that interns post their code in what ever state as draft PRs, at the end of the internship so as not to lose any progress.

dragon540:
As Jeff stated, it was mainly intended to save the current progress at the time of the end of mentorship. However, I will try to fix the issues this weekend and get the test bench readied now.

PR #247, opened 2025-03-10
Added Vector 1.0 Uop Generation Support

kathlenemagnus:
Olympia now supports uop generation for all vector instructions EXCEPT for some permutes (e.g. gathers, compresses, some complex slides). The uop generation for permutes is very dependent on how their execution unit is designed, so I've left them unimplemented for now. All types of uop generation have a test.


arches/isa_json/gen_uarch_rv64v_json.py
```diff
@@ -13,44 +13,109 @@
...
-    "vsetvl" :     {"pipe" : "vset", "latency" : 1},
-    "vsetivli" :   {"pipe" : "vset", "latency" : 1},
-
-# TODO: Vector Loads and Stores: Vector Unit-Stride Instructions
-# TODO: Vector Loads and Stores: Vector Strided Instructions
-# TODO: Vector Loads and Stores: Vector Indexed Instructions
-# TODO: Vector Loads and Stores: Unit-stride Fault-Only-First Loads
-# TODO: Vector Loads and Stores: Vector Load/Store Segment Instructions
-# TODO: Vector Loads and Stores: Vector Load/Store Whole Register Instructions
+    "vsetvli"  : {"pipe" : "vset", "latency" : 1},
```
aarongchan, line 16: Not sure how many more vector instructions would need to be added for this script, but for code readability would it be better to just make this it's own separate json file?

kathlenemagnus: I could create some header files that break things up a bit. I'll keep that in mind when we work on the vector load store instructions.

klingaard (approved):

PR #249, opened 2025-03-23
Rename revamp

klingaard:
Working with the previous Rename code base illustrated a series design issues that I needed to address.

In this PR, I've changed Rename to support the following:

- Move elimination ([example](https://easyperf.net/blog/2018/04/22/What-optimizations-you-can-expect-from-CPU#move-elimination))
- Partial vs full Renaming -- should rename make sure it can rename ALL instructions before renaming any of them?
- Faster resource availability calculation
- A more advanced RenameData structure
- Added ROB-targeted instruction support -- instructions can can "executed" at rename and sent directly to the ROB
- Code cleanup including a run through clang-format for those files that I touched

Keeping this as a draft for now as I need to update documentation and testing.

### DIFF
--- a/core/rename/Rename.cpp
+++ b/core/rename/Rename.cpp
@@ -1,13 +1,15 @@
+// <Rename.cpp> -*- C++ -*-
+
 /**
- * @file   Rename.cpp
- * @brief
- *
- *
+ * @file  Rename.cpp
+ * @brief Implementation of a simple rename block
  */
 
 #include <algorithm>
+#include <ranges>
+
 #include "CoreUtils.hpp"
-#include "rename/Rename.hpp"
+#include "Rename.hpp"
 #include "sparta/events/StartupEvent.hpp"
 #include "sparta/app/FeatureConfiguration.hpp"
 #include "sparta/report/DatabaseInterface.hpp"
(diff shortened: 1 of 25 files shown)


core/rename/Rename.cpp
```diff
@@ -137,398 +184,612 @@ namespace olympia
...
-
-                const auto & srcs = renaming_inst->getSourceOpInfoList();
-                const auto & dests = renaming_inst->getDestOpInfoList();
-                for (const auto & src : srcs)
+                auto & rcomp = regfile_components_[reg_file];
+                auto & freelist = rcomp.freelist;
+
+                sparta_assert(!freelist.empty(), "Freelist should never be empty");
+
+                uint32_t prf = 0;
```
aarongchan, line 600: Would it be better to initialize `prf` to `std::numeric_limits<uint32_t>::max()` and add an assert, similar to what you did in `RenameData` struct? I know logically the code currently always sets the PRF value based on freelist or move elimination, but in case someone makes a future change to move elimination, an easy bug could come up where the PRF value is never set and we default to always setting the prf to 0.

klingaard: Can do!

core/rename/Rename.cpp
```diff
@@ -137,398 +183,616 @@ namespace olympia
...
+                    if (false == dest.is_x0)
                     {
-                        --reference_counter_[data_reg.rf][data_reg.val];
-                        if (reference_counter_[data_reg.rf][data_reg.val] <= 0)
-                        {
-                            // freeing data register value, because it's not in the source list, so
-                            // won't get caught below
-                            freelist_[data_reg.rf].push(data_reg.val);
-                        }
+                        ++uop_queue_regcount_data_.cumulative_reg_counts[reg_file];
```
jeffnye-gh, line 347: this is a question, in the case where: _the freelist is empty, rename does not need to stall if the instruction has no dest,_ 

it seems like this case is handled naturally by this code + mavis. Is that correct? 

I spot checked a couple no-dest instructions with mavis and it returns a 0 vector for dest.

docs/rename.adoc
```diff
@@ -0,0 +1,356 @@
...
+instruction's sources must be renamed _first_ before the instruction's
+destinations.  If this order is not followed and an instruction's
+destination is the same as any of it's sources the machine will
+deadlock.
+
+===== Renaming Sources
+
+Rename maintains a map of current PRFs that the latest values for a
+given ARF (`map_table_`). During source renaming, the instruction's
+ARF sources are "replaced" by the current PRF sources being written by
```
jeffnye-gh, line 201: is there a word missing?
"Rename maintains a map of current PRFs that **hold** the latest values for a given ARF"

jeffnye-gh (approved):
looks good to me, i did not see any thing to comment on in the code. 

I did have a question about no-dest instructions.

your doc is great, btw.  trivial typo comment.

core/rename/Rename.cpp
klingaard: You are correct.  If a group of instructions from Decode contains only source-reading instructions, and the freelilst is empty, then Rename can and will rename that group.

docs/rename.adoc
klingaard: Thanks for catching that.  Fixed.

Issue #254, opened 2025-05-02
Model terminates (Run successful) before completion of trace


rajatbhatia1:
Found an interesting case where the model stopped making forward progress and terminated indicating a complete run.
I tried to debug and found that the STF read did not reach the end of trace, but the simulator ran out of events and thought everything was done.
Further debug revealed that a Flush cleared the fetch unit, but downstream pipes were also empty, and there were no events that started fetch unit to start fetching instructions. 

Possible solution (it seems to work fine) is to schedule fetch event in next cycle when fetch unit handles the flush.

rajatbhatia1:
Please assign to me

rajatbhatia1:
Pushed the fix:
https://github.com/riscv-software-src/riscv-perf-model/pull/256

PR #256, opened 2025-05-23
Fix for Fetch Hang (#254)

rajatbhatia1:
It was observed that in certain long tests olympia model completed prematurely.

This is because fetch unit is not restarted after flush and in this particular case, there were no credits or other events that triggered fetch to start fetching instructions

Solution is to schedule fetching in next cycle after flush, just as done on initialization

### DIFF
--- a/core/fetch/Fetch.cpp
+++ b/core/fetch/Fetch.cpp
@@ -260,6 +260,11 @@ namespace olympia
 
         // No longer speculative
         // speculative_path_ = false;
+
+	// Fix for Issue #254
+	// It is possible that we do not have any external trigger to restart
+	// fetch, so force bootstrap it like when initialized
+	ev_fetch_insts->schedule(1);
     }
 
     void Fetch::onROBTerminate_(const bool & stopped)


aarongchan:
Fix looks good, I was worried that we could run into a case where the fetch event gets scheduled twice if somehow a credit or event triggered it as well as we trigger it from fetch clear. However, based on the documentation for `SingleCycleUniqueEvent`, that doesn't seem to be an issue.
From the MAP [documentation](https://sparcians.github.io/map/classsparta_1_1SingleCycleUniqueEvent.html):
"This type of [Event](https://sparcians.github.io/map/classsparta_1_1Event.html) is exactly like [sparta::Event](https://sparcians.github.io/map/classsparta_1_1Event.html), but a [SingleCycleUniqueEvent](https://sparcians.github.io/map/classsparta_1_1SingleCycleUniqueEvent.html) will not schedule itself twice on the scheduler for the same time and will only schedule itself on the subsequent cycle."

@klingaard please correct me if I'm wrong on interpreting this.

klingaard:
> @klingaard please correct me if I'm wrong on interpreting this.

No correction needed.  That's how it works.

klingaard (approved):

Issue #253, opened 2025-04-30
SYS instructions flushing from ROB do not prevent further retires in same cycle causing program order assertion


rajatbhatia1:
I have identified a bug in the model where SYS instructions flushing from ROB do not prevent further retires in same cycle causing program order assertion

Here is the test case
[sys_flush_test.json](https://github.com/user-attachments/files/19985045/sys_flush_test.json)

Command line:
./olympia --arch ../arches/big_core.yaml ./sys_flush_test.json

Exception while running
ex_inst.getProgramID() == expected_program_id_: 
Unexpected program ID when retiring instruction
(suggests wrong program order)
 expected: 15
 received: 11
 UID: 235
 incr: 1
 inst uid:98    RETIRED 0 pid:11 uopid:0 'csrrs	18,0' : in file: '/work/rbhatia2/repos/olympia/github-riscv-software-src-riscv-perf-model/core/ROB.cpp', on line: 153

rajatbhatia1:
I have the solution for this issue. Please assign to me

PR #257, opened 2025-05-24
Flushing SYS instructions on retire should halt same cycle retires

rajatbhatia1:
Failing test and solution for Issue #253 

Some of the SYS instrutions are executed on retire and may cause a flush. Olympia currently allows the youger uops to retire in the same cycle. 

However, when the flushing sys uop is retired, the program order check fails, as younger uops have retired out of order.

Solution is to stop retiring uops when this happens.


klingaard (approved):
Appreciated the fix and the test!

PR #264, opened 2025-07-11
Docker workflow for STF trace generation

vinicius-r-silva:
This PR introduces a new folder, docker_stf_trace_gen, as part of the ongoing Workload Analysis Flow mentorship. It includes:
* A Dockerfile that builds a container with the RISC-V GNU toolchain, pk, and Spike-STF.
* Supporting scripts for building and running the Docker image
* Documentation Readme

The current workflow does not yet include the commands to produce STF files. These will be added in a future PR. For now it only includes the basis for the documentation and the Dockerfile with the necessary tools.

### DIFF
--- a/traces/docker_stf_trace_gen/README.md
+++ b/traces/docker_stf_trace_gen/README.md
@@ -1,0 +1,58 @@
+# Spike-STF trace generation
+
+> This workflow is still under development. For the most up-to-date instructions, refer to this [README](../README.md)
+
+## Table of Contents
+
+1. [Introduction](#introduction)
+1. [Dependencies](#dependencies)
+1. [Project Structure](#project-structure)
+1. [Usage](#usage)
+    1. [Build the Docker Image](#build-the-docker-image)
+    1. [Run STF Trace Generation](#run-stf-trace-generation)
+
+## Introduction
+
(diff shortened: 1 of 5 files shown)


traces/docker_stf_trace_gen/README.md
```diff
@@ -0,0 +1,58 @@
+# Spike-STF trace generation
+
+> This workflow is still under development. For the most up-to-date instructions, refer to this [README](../README.md)
```
klingaard, line 3: Which README is "this"?

klingaard (commented):
LGTM as a great first start.  Will let Arup/Jaksa approve

traces/docker_stf_trace_gen/README.md
vinicius-r-silva: Is the [Generating Input for Olympia
](https://github.com/riscv-software-src/riscv-perf-model/blob/master/traces/README.md)

I’ve updated the text to be more explicitly and avoid using the word "this"
> For the most up-to-date instructions, refer to the [Generating Input for Olympia](https://github.com/riscv-software-src/riscv-perf-model/blob/master/traces/README.md) README

arupc (approved):

Issue #268, opened 2025-07-26
Add parameter to top-level model (App-level) to select topology


oms-vmicro:
Looking at `main.cpp`, it seems like there may have been an intention to permit the use of different topologies besides the default "Simple" topology (See https://github.com/riscv-software-src/riscv-perf-model/blob/master/sim/main.cpp#L101).

Simple potential change if it would be of use / help extending the model: add knob at "app" level to select topology and then pass topology into OlympiaSim constructor instead of hard-coded "simple".  Will default this parameter to "simple" so functionally it will be the same, but will then permit generic infrastructure to select topology.  Though of course, one has to go and then specify that topology.

oms-vmicro:
PR: https://github.com/riscv-software-src/riscv-perf-model/pull/274

This approach makes topology a parameter within the CPU unit itself.

PR #274, opened 2025-08-14
Add topology parameter to CPU unit. (#268)

oms-vmicro:
This moves `topology` into a CPU-level parameter and explicitly sets the default in the small_core config to the "simple" topology.  It also removes topology from the `OlympiaSim` constructor.


klingaard (approved):
This looks good.  Thanks for doing that!

oms-vmicro:
Had not run `make regress` with the last push.  New push updates test file `OlympiaSim` constructor calls to remove the deprecated topology parameter at that level (superseded by new CPU-level topology parameter).

Regression now builds and passes all tests and hopefully should pass CI/CD as well.

PR #278, opened 2025-09-17
Adding the End to End flow

Jatin-exe:
Implemented:
- adds slicing the workloads using the simpointed output 
-  builds olympia inside docker. 
-  refactor for DockerOrchesterator (host) and run_command on container
-  unfiy redundant functions and remove redundant code
- adds outputs dir to .gitignore (the output generated from the flow)
- update yaml spec to include diverse workloads without code changes
- use an obj file as input for the flow rather than source code.

To do:
- test qemu simpoint generation plugin
- performance modelling using olympia
- more testing with 
- unify with trace-archive


arupc-vmicro:
@Jatin-exe Please provide an ETA to address the review comments.

PR #280, opened 2025-10-17
Move Mavis forward; use Boost::json

klingaard:
Moved to _almost_ the tip of Mavis.  The head of Mavis requires c++23 features (@bdutro) which we might not be ready for yet.

Started to deprecate the use of nlohmann::json and moved to Boost::json.  Can't remove it completely as FSL still uses it and we're pointing to a mirror (for now).  Put that header in `deprecated` directory for now.

Also cleaned up some header file allocations to improve compile time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.19)
-set (CMAKE_CXX_STANDARD 17)
+set (CMAKE_CXX_STANDARD 20)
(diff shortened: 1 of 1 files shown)


bdutro:
Mavis should compile cleanly with C++20 unless you're using a Clang version older than 16 (this branch has a preliminary fix for that: https://github.com/sparcians/mavis/tree/dev/bdutro/clang-15-std-views-fix).

PR #284, opened 2025-12-29
feat: Add git SHA to simulator output for reproducibility(#103)

### LINKED ISSUE #103: Add git SHA to output
Can we modify Olympia such that it always outputs the git SHA used to build the simulator? This will be helpful for reproducibility.

16srivarshitha:
## Description

Adds git commit SHA to the Olympia simulator output to improve reproducibility and traceability of simulation results.

## Motivation

This enhancement addresses the need for better reproducibility by clearly indicating the exact version of the code used to build the simulator. This is particularly valuable for:

- **Bug reports**: Maintainers can identify the exact code version being used
- **Performance comparisons**: Researchers can verify they're comparing equivalent versions
- **Research reproducibility**: Results can be traced back to specific commits
- **Development tracking**: The `-dirty` flag helps identify when local modifications are present

Resolves #103

## Changes

**Build System:**
- Added `GenerateGitVersion.cmake` script that captures Git SHA and dirty status at build time (not configure time)
- Implemented generated header approach (`git_version.h`) with proper dependency tracking on `.git/HEAD` and `.git/index`
- This ensures the version info automatically updates whenever Git state changes, without requiring manual CMake reconfiguration
- Modified `core/CMakeLists.txt` to generate the header before compilation and made include directories `PUBLIC` to allow access from `sim/` directory

**Code Integration:**
- Modified `sim/main.cpp` to populate the `reproduction_info` field in `sparta::SimulationInfo` instead of printing directly to stdout (as suggested by maintainers)
- The Git SHA is now part of the structured simulation metadata

**CI/CD:**
- Added `workflow_dispatch` trigger to Ubuntu build workflow to enable manual regression testing

## Testing

- Code changes are syntactically correct and follow CMake/C++ best practices
- Changes are minimal, localized, and non-breaking
- Verified `-dirty` flag correctly identifies uncommitted changes
- Successfully validated build and full regression suite via GitHub Actions on forked repository
- Fallback to "unknown" if git is unavailable

## Example Output

After this change, the simulator displays:
```
# Sparta Version: x.x.x
# Git SHA: a1b2c3d
```

Or if there are uncommitted changes:
```
# Sparta Version: x.x.x
# Git SHA: a1b2c3d-dirty
```

### DIFF
--- a/sim/main.cpp
+++ b/sim/main.cpp
@@ -37,8 +37,10 @@ int main(int argc, char **argv)
     const std::string olympia_version = " " + std::string(OLYMPIA_VERSION);
     const std::string reproduction_info = "Git SHA: " + std::string(GIT_COMMIT_HASH);
     sparta::SimulationInfo::getInstance() = sparta::SimulationInfo("Olympia RISC-V Perf Model ",
-                                                                argc, argv, olympia_version.c_str(),
-                                                            reproduction_info.c_str(), {});
+                                                                    argc, 
+                                                                    argv, 
+                                                                    olympia_version.c_str(),
+                                                                    reproduction_info.c_str(), {});
     const bool show_field_names = true;
     sparta::SimulationInfo::getInstance().write(std::cout, "# ", "\n", show_field_names);
     std::cout << "# Sparta Version: " << sparta::SimulationInfo::sparta_version << std::endl;


sim/main.cpp
```diff
@@ -35,12 +35,18 @@ int main(int argc, char **argv)
 
     const std::string olympia_version = " " + std::string(OLYMPIA_VERSION);
     sparta::SimulationInfo::getInstance() = sparta::SimulationInfo("Olympia RISC-V Perf Model ",
-                                                                   argc, argv, olympia_version.c_str(),
-                                                                   "", {});
+                                                                argc, argv, olympia_version.c_str(),
+                                                                "", {});
```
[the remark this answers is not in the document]
16srivarshitha: Thanks for the review and kind words!
You’re absolutely right about the dependency issue. At the moment, the git SHA is captured at build time, so if files are modified without triggering a rebuild of `main.cpp`, the SHA won’t update.
I think the best fix is to move the Git capture into a generated header (e.g., `git_version.h`). Using `add_custom_command`, I can check the git state on every build and have `main.cpp` include this header, ensuring it recompiles whenever the SHA or dirty status changes.

About the SimulationInfo - I see what you mean about the placeholder. I’ll populate it with the git SHA instead of printing to stdout.
I'll work on these changes and update the PR. Let me know if you’d like any additional adjustments.

16srivarshitha:
@klingaard Thanks again for the detailed feedback! I've updated the implementation to address both concerns you raised.

**Changes made:**

1. **Dependency tracking**: Switched from `execute_process` (which runs at configure time) to a generated header approach using `add_custom_command`. The header generation now depends on `.git/HEAD` and `.git/index`, so any git state changes automatically trigger a rebuild of `main.cpp` - no need to manually reconfigure CMake.

2. **SimulationInfo integration**: You're absolutely right about using the framework properly. I've moved the Git SHA into the `reproduction_info` field of `sparta::SimulationInfo` instead of printing it separately. Much cleaner this way!

I also had to set the `core` library's include directories to `PUBLIC` to make the generated header accessible from the `sim/` directory - ran into some build issues otherwise.

The regression tests are all passing on my fork: https://github.com/16srivarshitha/riscv-perf-model/actions.

Feel free to check the workflow runs if you want to verify before pulling this in! Let me know if this approach looks good or if you'd like me to adjust anything!

sim/main.cpp
```diff
@@ -34,9 +35,10 @@ int main(int argc, char **argv)
     DEFAULTS.arch_search_dirs = {"arches"}; // Where --arch will be resolved by default
 
     const std::string olympia_version = " " + std::string(OLYMPIA_VERSION);
+    const std::string reproduction_info = "Git SHA: " + std::string(GIT_COMMIT_HASH);
     sparta::SimulationInfo::getInstance() = sparta::SimulationInfo("Olympia RISC-V Perf Model ",
-                                                                   argc, argv, olympia_version.c_str(),
-                                                                   "", {});
+                                                                argc, argv, olympia_version.c_str(),
+                                                            reproduction_info.c_str(), {});
```
klingaard: Do you mind fixing the spacing so it lines up?

16srivarshitha: Fixed the spacing! I've aligned the arguments vertically as requested. Thanks for catching that. Ready for final review.

PR #294, opened 2026-02-06
Add execution-driven CPI attribution framework

sagar-h007:
### What this PR is about

This PR adds a CPI (Cycles Per Instruction) attribution framework to Olympia so it’s easier to understand where execution time is actually being spent in the pipeline.

Until now, we had IPC and a few aggregate stall counters, but no clear way to tie stalls back to retiring instructions. This change fills that gap by tracking CPI at the instruction level and aggregating it at retirement.

⸻

### What’s included
- Per-instruction CPI breakdown across 8 pipeline stages:
- Fetch, Decode, Rename, Dispatch, Issue Queue, Execute, Memory, and ROB
- Lightweight timestamping as instructions move through the pipeline
- CPI calculation done at retirement, so there’s no added pressure on the execution path
- ROB-level aggregation using Sparta counters, which shows up naturally in reports

⸻

### How it works (high level)

Each pipeline stage records when an instruction enters it.
When the instruction retires, CPI components are computed as differences between these timestamps and accumulated into ROB counters.

This keeps the implementation simple, execution-driven, and easy to extend.

⸻

### Design notes
- Attribution is stage-based, not fine-grained causal (e.g., L1 vs L2 vs TLB)
- That tradeoff is intentional for a first step and keeps the model stable
- More detailed breakdowns can be layered on later if needed

⸻

### Why this is useful

This makes it possible to answer questions like:
- Are we front-end bound, memory bound, or execution bound?
- Which pipeline stages dominate CPI for a given workload?
- How do micro-architectural changes impact real execution time?

⸻

### Scope
- New CPI/timestamp metadata on instructions
- New CPI finalization logic
- Small instrumentation changes across Fetch → ROB
- ROB-level CPI aggregation via Sparta

⸻

### Fixes: #293 

If this direction looks good, I’m happy to iterate on naming, attribution granularity, or validation strategy based on feedback.


klingaard (changes requested):

core/Inst_cpi.cpp
klingaard: Probably don't need a separate file for this. Go ahead and put this code in Inst.cpp.

core/Inst_cpi.cpp
```diff
@@ -0,0 +1,56 @@
...
+        // Calculate cycle durations for each stage based on timestamps
+        // Note: For simplicity, we attribute all cycles between stage transitions
+        // to the appropriate category. More sophisticated attribution could
+        // differentiate between productive work and stalls within a stage.
+        
+        auto& breakdown = cpi_breakdown_;
+        const auto& ts = timestamps_;
+        
+        // Fetch stalls: From fetch to decode
+        if (ts.decode_enter > ts.fetch_enter) {
```
klingaard: Shouldn't these `if` statement really be asserts?

core/Inst.hpp
```diff
@@ -69,6 +69,31 @@ namespace olympia
...
+            uint64_t memory_stall_cycles = 0;         //! TLB miss, cache miss, store wait
+            uint64_t rob_stall_cycles = 0;            //! Waiting to become oldest
+            
+            //! Get total cycles across all categories
+            uint64_t getTotalCycles() const {
+                return fetch_stall_cycles + decode_stall_cycles + rename_stall_cycles +
+                       dispatch_stall_cycles + issue_queue_stall_cycles + execute_cycles +
+                       memory_stall_cycles + rob_stall_cycles;
+            }
+        };
```
klingaard: This can be collapsed into an array of counters as well.  The array should indexable on the status enum

sagar-h007: using an array indexed by the status enum would definitely simplify iteration and make the structure easier to extend if more categories are added later.

I initially kept them as named fields for readability while validating the attribution logic, but I agree an indexed container would be cleaner long-term. I can refactor this into an array with an enum index mapping if you’d prefer that approach.

core/Inst.hpp
```diff
@@ -69,6 +69,41 @@ namespace olympia
...
+        //! Tracks cycles attributed to different microarchitectural events
+        struct CPIBreakdown
+        {
+            std::array<uint64_t, static_cast<size_t>(CPIStallType::NUM_TYPES)> cycles = {0};
+
+            //! Get total cycles across all categories
+            uint64_t getTotalCycles() const
+            {
+                uint64_t total = 0;
+                for (const auto c : cycles)
```
klingaard, line 93: Consider `std::accumulate`

core/ROB.cpp
```diff
@@ -19,13 +19,37 @@ namespace olympia
...
+                                "Total cycles waiting for operands in issue queue",
+                                sparta::Counter::COUNT_NORMAL),
+        cpi_execute_cycles_(&unit_stat_set_, "cpi_execute_cycles",
+                            "Total cycles in execution across all instructions",
+                            sparta::Counter::COUNT_NORMAL),
+        cpi_memory_stalls_(&unit_stat_set_, "cpi_memory_stalls",
+                           "Total cycles stalled on memory operations",
+                           sparta::Counter::COUNT_NORMAL),
+        cpi_rob_stalls_(&unit_stat_set_, "cpi_rob_stalls", "Total cycles waiting in ROB",
+                        sparta::Counter::COUNT_NORMAL),
```
klingaard, line 52: You can put these into a `vector` of counters based on the enum values for these counters.  `sparta::Counters` are movable components.

core/ROB.cpp
```diff
@@ -169,6 +193,23 @@ namespace olympia
...
+                    breakdown.cycles[static_cast<size_t>(Inst::CPIStallType::FETCH)];
+                cpi_decode_stalls_ +=
+                    breakdown.cycles[static_cast<size_t>(Inst::CPIStallType::DECODE)];
+                cpi_rename_stalls_ +=
+                    breakdown.cycles[static_cast<size_t>(Inst::CPIStallType::RENAME)];
+                cpi_dispatch_stalls_ +=
+                    breakdown.cycles[static_cast<size_t>(Inst::CPIStallType::DISPATCH)];
+                cpi_execute_cycles_ +=
+                    breakdown.cycles[static_cast<size_t>(Inst::CPIStallType::EXECUTE)];
+                cpi_rob_stalls_ += breakdown.cycles[static_cast<size_t>(Inst::CPIStallType::ROB)];
```
klingaard, line 211: Can you not iterate over the breakdown?  If some of the counters are zero, then who cares, right?

klingaard (commented):
What's your status on completing this work?

klingaard:
@sagar-h007 what's the status of the PR?  Do you think you'll be able to get back to it?

PR #301, opened 2026-04-20
Add enhanced branch predictor phase-1 integration and tests

kumavatkomal:
This is a PR for Phase-1 branch predictor integration work.
The goal is to land implementation and test wiring first, while keeping EDM comparison and final performance claims for follow-up work.

**Included in this PR**

- Enhanced branch predictor integration in fetch flow
- Predictor selection and config wiring
- Branch prediction evaluation and update path integration
- Branch predictor unit and integration test updates
- Required architecture parameter updates for predictor selection

**Not included yet**

- Final EDM vs trace comparison results
- Final measured claim locking based on EDM runs
- Final report artifacts and final comparison tables

**Validation completed**

- Build completed for updated targets
- Branch predictor tests passed after integration updates

**Commands used for validation**

- cmake --build build --target fetch BranchPred_test -j4
- ./build/test/core/branch_pred/BranchPred_test

**issue tracking**

Related to riscv-software-src/riscv-perf-model#1


core/fetch/Fetch.cpp
```diff
@@ -55,6 +61,39 @@ namespace olympia
         node->getParent()->registerForNotification<bool, Fetch, &Fetch::onROBTerminate_>(
             this, "rob_stopped_notif_channel", false /* ROB maybe not be constructed yet */);
 
+        std::transform(branch_predictor_name_.begin(), branch_predictor_name_.end(),
+                       branch_predictor_name_.begin(),
+                       [](unsigned char ch) { return static_cast<char>(std::tolower(ch)); });
```
bdutro: `std::transform` shouldn't be used for in-place modification of a container:

> std::transform does not guarantee in-order application of unary_op or binary_op. To apply a function to a sequence in-order or to apply a function that modifies the elements of a sequence, use std::for_each.

core/fetch/Fetch.cpp
```diff
@@ -55,6 +61,39 @@ namespace olympia
...
+
+        const auto is_power_of_two = [](const uint32_t value) {
+            return value && ((value & (value - 1)) == 0);
+        };
```
bdutro: Use `sparta::utils::is_power_of_2`

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,163 @@
...
+ *   - Training: once the branch resolves, score the pending prediction,
+ *     update BHT direction state, then refresh BTB metadata.
+ */
+
+namespace
+{
+    bool isPowerOfTwo(const uint32_t value)
+    {
+        return value && ((value & (value - 1)) == 0);
+    }
```
bdutro: Use `sparta::utils::is_power_of_2`

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,163 @@
...
+            kBTBCacheLineBytes,
+            kBTBCacheLineBytes,
+            BTBCacheLine(),
+            *btb_replacement_policy_);
+    }
+
+    uint64_t EnhancedBranchPredictor::btbCacheAddress_(uint64_t fetch_pc) const {
+        // Convert instruction-addressed fetch PC into the cache-line address
+        // expected by SimpleCache2 while preserving intended BTB indexing.
+        return fetch_pc << (kBTBCacheLineShift - kCompressedIndexShift);
```
bdutro, line 51: I think you want this to be a right-shift?

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,163 @@
...
+        if ((cache_line != nullptr) && cache_line->isValid()) {
+            if (entry != nullptr) {
+                *entry = cache_line->btb_entry;
+            }
+            // Keep replacement state hot for recently used BTB lines.
+            btb_cache_->touchMRU(*cache_line);
+            return true;
+        }
+        return false;
+    }
```
bdutro: You could just have this return a pointer to `cache_line->btb_entry` on a hit and a nullptr on a miss. This would also avoid the overhead of copying around a `BTBEntry` on every hit.

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    void EnhancedBranchPredictor::btbUpdate_(uint64_t fetch_pc, const BTBEntry & entry) {
+        const auto btb_addr = btbCacheAddress_(fetch_pc);
+        // One path for both invalid-line fill and victim replacement.
+        auto & replacement_line = btb_cache_->getLineForReplacementWithInvalidCheck(btb_addr);
+        btb_cache_->allocateWithMRUUpdate(replacement_line, btb_addr);
+        replacement_line.btb_entry = entry;
```
bdutro: Consider altering `replacement_line.btb_entry` directly instead of copying over it.

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,163 @@
...
+            entry.branch_idx = update.branch_idx;
+            if (update.actually_taken) {
+                entry.predicted_PC = update.corrected_PC;
+            }
+        } else {
+            const uint64_t default_target = update.actually_taken ?
+                update.corrected_PC :
+                (update.fetch_PC + update.branch_idx + bytes_per_inst);
+            entry = BTBEntry(update.branch_idx, default_target);
+        }
```
bdutro: One potential optimization here is to only allocate new BTB entries after a taken branch is mispredicted.

core/fetch/Fetch.cpp
```diff
@@ -188,6 +231,74 @@ namespace olympia
...
+            found_branch = true;
+            actual_branch_idx = i;
+            actual_branch_inst = inst;
+            if (inst->isTakenBranch()) {
+                actual_next_pc = inst->getTargetVAddr();
+            } else {
+                actual_next_pc = inst->getPC() + BranchPredictorIFType::bytes_per_inst;
+            }
+            break;
+        }
```
bdutro: Alternatively, you could do something like this:

```
auto it = std::find_if(insts_to_send->begin(), insts_to_send->end(), [](const InstPtr& inst) { return inst->isBranch(); });

if(it == insts_to_send->end())
{
    return;
}

// Now process the branch update
```

core/fetch/Fetch.cpp
```diff
@@ -188,6 +231,74 @@ namespace olympia
...
+        input.fetch_PC = (*insts_to_send->begin())->getPC();
+        const auto prediction = branch_predictor_->getPrediction(input);
+
+        bool found_branch = false;
+        uint32_t actual_branch_idx = num_insts_to_fetch_;
+        uint64_t actual_next_pc = input.fetch_PC + num_insts_to_fetch_ * BranchPredictorIFType::bytes_per_inst;
+        InstPtr actual_branch_inst;
+
+        // Predictor contract: one prediction per fetch packet, keyed by fetch PC.
+        // We therefore train/evaluate against the first branch in this packet.
```
bdutro, line 240: What if there are multiple branches in a fetch packet? If there is a never taken conditional branch followed by an always taken conditional branch, the second branch will never be inserted into the BTB and we will always mispredict.

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,163 @@
...
+        prediction.branch_idx = max_fetch_insts_;
+        prediction.predicted_PC = input.fetch_PC + max_fetch_insts_ * bytes_per_inst;
+
+        BTBEntry btb_entry;
+        if (btbLookup_(input.fetch_PC, &btb_entry)) {
+            prediction.branch_idx = btb_entry.branch_idx;
+            if (bhtPredict_(input.fetch_PC)) {
+                prediction.predicted_PC = btb_entry.predicted_PC;
+            } else {
+                prediction.predicted_PC = input.fetch_PC + prediction.branch_idx + bytes_per_inst;
```
bdutro, line 102: This isn't the correct fallthrough PC. The way `branch_idx` is calculated in `Fetch::evaluateBranchPrediction_` doesn't account for non-branch instructions in a fetch bundle.

arches/medium_core.yaml
```diff
@@ -7,6 +7,7 @@ include: small_core.yaml
 
 top.cpu.core0:
   fetch.params.num_to_fetch:   3
+  fetch.params.branch_predictor: simple
```
klingaard: Since medium core inherits from small core, it automatically picks up the simple branch predictor.

arches/big_core.yaml
```diff
@@ -7,6 +7,7 @@ include: medium_core.yaml
 
 top.cpu.core0:
   fetch.params.num_to_fetch:   8
+  fetch.params.branch_predictor: simple
```
klingaard: Since big core inherits from medium/small core, it automatically picks up the simple branch predictor.

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    uint64_t EnhancedBranchPredictor::btbCacheAddress_(uint64_t fetch_pc) const {
+        // Convert instruction-addressed fetch PC into the cache-line address
+        // expected by SimpleCache2 while preserving intended BTB indexing.
+        return fetch_pc << (kBTBCacheLineShift - kCompressedIndexShift);
+    }
+
+    uint32_t EnhancedBranchPredictor::bhtIndex_(uint64_t fetch_pc) const {
+        return static_cast<uint32_t>((fetch_pc >> 1) & (bht_entries_ - 1));
```
klingaard: Won't this lose bit precision?

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,153 @@
...
+        // One path for both invalid-line fill and victim replacement.
+        auto & replacement_line = btb_cache_->getLineForReplacementWithInvalidCheck(btb_addr);
+        btb_cache_->allocateWithMRUUpdate(replacement_line, btb_addr);
+        // Directly modify the entry in place
+        replacement_line.btb_entry.branch_idx = entry.branch_idx;
+        replacement_line.btb_entry.predicted_PC = entry.predicted_PC;
+    }
+
+    bool EnhancedBranchPredictor::bhtPredict_(uint64_t fetch_pc) const {
+        return branch_history_table_[bhtIndex_(fetch_pc)] > 1;
```
klingaard: Suggest you use `at` to avoid overflow: `branch_history_table_.at()`

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    void EnhancedBranchPredictor::bhtUpdate_(uint64_t fetch_pc, bool actually_taken) {
+        auto & counter = branch_history_table_[bhtIndex_(fetch_pc)];
```
klingaard: Use `at()`

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+        BTBEntry* btb_entry = btbLookup_(input.fetch_PC);
+        if (btb_entry != nullptr) {
+            prediction.branch_idx = btb_entry->branch_idx;
+            if (bhtPredict_(input.fetch_PC)) {
+                prediction.predicted_PC = btb_entry->predicted_PC;
+            } else {
+                prediction.predicted_PC = input.fetch_PC + prediction.branch_idx + bytes_per_inst;
+            }
+        }
```
klingaard, line 104: Soooo... what happens if there isn't a BTB entry?  Do we still have a prediction?  When does a BTBEntry get allocated?

core/fetch/EnhancedBranchPred.cpp
```diff
@@ -0,0 +1,153 @@
...
+        btb_replacement_policy_ = olympia::ReplacementFactory::selectReplacementPolicy("LRU", btb_ways_);
+        btb_cache_ = std::make_unique<sparta::cache::SimpleCache2<BTBCacheLine>>(
+            btb_entries_,
+            kBTBCacheLineBytes,
+            kBTBCacheLineBytes,
+            BTBCacheLine(),
+            *btb_replacement_policy_);
+    }
+
+    uint64_t EnhancedBranchPredictor::btbCacheAddress_(uint64_t fetch_pc) const {
```
klingaard: I think this method can be a `constexpr`

core/fetch/Fetch.cpp
```diff
@@ -23,6 +25,10 @@ namespace olympia
         sparta::Unit(node),
         my_clk_(getClock()),
         num_insts_to_fetch_(p->num_to_fetch),
+        branch_predictor_name_(p->branch_predictor),
+        enhanced_btb_entries_(p->enhanced_btb_entries),
+        enhanced_btb_ways_(p->enhanced_btb_ways),
+        enhanced_bht_entries_(p->enhanced_bht_entries),
```
klingaard: Why are these members of Fetch?  Does Fetch need to know this information?  Looks like it's passed to the predictor instantiations below.

core/fetch/Fetch.cpp
```diff
@@ -55,6 +61,34 @@ namespace olympia
...
+        } else if (branch_predictor_name_ == "enhanced") {
+            sparta_assert(enhanced_btb_entries_ > 0, "enhanced_btb_entries must be > 0");
+            sparta_assert(enhanced_btb_ways_ > 0, "enhanced_btb_ways must be > 0");
+            sparta_assert(enhanced_bht_entries_ > 0, "enhanced_bht_entries must be > 0");
+            sparta_assert((enhanced_btb_entries_ % enhanced_btb_ways_) == 0,
+                          "enhanced_btb_entries must be divisible by enhanced_btb_ways");
+            sparta_assert(sparta::utils::is_power_of_2(enhanced_btb_entries_ / enhanced_btb_ways_),
+                          "enhanced BTB set count must be power-of-two");
+            sparta_assert(sparta::utils::is_power_of_2(enhanced_bht_entries_),
+                          "enhanced_bht_entries must be power-of-two");
```
klingaard: Should these asserts be in EnhancedBranchPredictor?

core/fetch/Fetch.cpp
```diff
@@ -188,6 +226,60 @@ namespace olympia
...
+        const uint64_t actual_next_pc = actual_branch_inst->isTakenBranch() ?
+            actual_branch_inst->getTargetVAddr() :
+            (actual_branch_inst->getPC() + BranchPredictorIFType::bytes_per_inst);
+
+        BranchPredictor::DefaultUpdate update;
+        update.fetch_PC = input.fetch_PC;
+        update.branch_idx = actual_branch_idx;
+        update.corrected_PC = actual_next_pc;
+        update.actually_taken = actual_branch_inst->isTakenBranch();
+        branch_predictor_->updatePredictor(update);
```
klingaard, line 253: I think the `updatePredictor` method should populate predictor object.  Fetch should know nothing about how the predictor works.

core/fetch/Fetch.cpp
```diff
@@ -188,6 +226,60 @@ namespace olympia
...
+
+        const bool mispredicted =
+            (prediction.branch_idx != actual_branch_idx) ||
+            (prediction.predicted_PC != actual_next_pc);
+
+        // Track end-to-end prediction quality at fetch integration level.
+        ++branch_predictions_;
+        if (mispredicted) {
```
klingaard, line 261: @bdutro question for you: should we count the mispredict here or when the branch is resolved?

core/fetch/Fetch.hpp
```diff
@@ -111,6 +121,21 @@ namespace olympia
...
+            BranchPredictor::DefaultInput>;
+        std::unique_ptr<BranchPredictorIFType> branch_predictor_;
+
+        // Kept for logging/debug visibility (normalized to lowercase in ctor).
+        std::string branch_predictor_name_;
+
+        // Geometry knobs consumed when enhanced predictor is selected.
+        const uint32_t enhanced_btb_entries_;
+        const uint32_t enhanced_btb_ways_;
+        const uint32_t enhanced_bht_entries_;
```
klingaard: Don't see a need to keep these around

klingaard (commented):
Love the testing!

core/fetch/Fetch.cpp
bdutro: We should count them when the branch is resolved to avoid overcounting mispredicts that happen in the shadow of another mispredict/flush/etc.

core/fetch/EnhancedBranchPred.cpp
kumavatkomal: The cast happens after masking, so precision is preserved. The mask `(bht_entries_ - 1)` extracts only the lower 9 bits (for 512 entries), which easily fits in uint32_t. No bits are lost.

Is there a specific precision concern I should address?

core/fetch/EnhancedBranchPred.cpp
kumavatkomal:
```cpp
// BTB Allocation Policy:
// - On BTB hit: Always update the existing entry with latest branch metadata
// - On BTB miss: Only allocate for taken branches (optimization to reduce pollution)
// - Not-taken branches don't allocate BTB entries (they follow straight-line fetch)
BTBEntry* existing_entry = btbLookup_(update.fetch_PC);

if (existing_entry != nullptr) {
    // Update existing BTB entry
    existing_entry->branch_idx = update.branch_idx;
    if (update.actually_taken) {
        existing_entry->predicted_PC = update.corrected_PC;
    }
} else if (update.actually_taken) {
    // Allocate new BTB entry only for taken branches
    BTBEntry new_entry(update.branch_idx, update.corrected_PC);
    btbUpdate_(update.fetch_PC, new_entry);
}
```
On BTB miss, we predict straight-line fetch. Allocation happens only for taken branches to reduce pollution. Let me know if it is correct or not.

core/fetch/Fetch.cpp
kumavatkomal: You're right - Fetch doesn't need to store these after construction. They're only used to pass to the predictor. Should I update this to read from ParameterSet and pass directly without storing as member variables?

core/fetch/Fetch.cpp
kumavatkomal: Fetch builds DefaultUpdate because it has the instruction data. This is the same pattern used by SimpleBranchPredictor - both receive DefaultUpdate objects. Should I change both predictors to use a different approach, or is the current design acceptable?

core/fetch/Fetch.hpp
kumavatkomal: Fixed - removed unused counter variables.

For the config variables (enhanced_btb_entries_, enhanced_btb_ways_, enhanced_bht_entries_): these are currently stored because they're read from ParameterSet and passed to predictor constructor. Should I update this to read and pass directly without storing? Let me know and I'll make the change.

bdutro: > Should I update this to read and pass directly without storing?

Yes, that's the preferred way to handle this.

core/fetch/Fetch.cpp
bdutro: No, let's keep the current interface for now. In the future we may want the update object to contain the instruction data.

core/fetch/EnhancedBranchPred.cpp
bdutro: Since this goes directly into a vector lookup, you could return a `size_t` instead and avoid a conversion.

core/fetch/EnhancedBranchPred.cpp
bdutro: That sounds good to me.

core/fetch/Fetch.cpp
```diff
@@ -55,6 +58,24 @@ namespace olympia
...
+
+        // Keep predictor selection explicit here so config errors fail fast.
+        if (branch_predictor_name_ == "simple") {
+            branch_predictor_.reset(new BranchPredictor::SimpleBranchPredictor(num_insts_to_fetch_));
+        } else if (branch_predictor_name_ == "enhanced") {
+            branch_predictor_.reset(new BranchPredictor::EnhancedBranchPredictor(
+                num_insts_to_fetch_,
+                p->enhanced_btb_entries,
+                p->enhanced_btb_ways,
+                p->enhanced_bht_entries));
```
bdutro: @kumavatkomal, you need to call `ignore()` on these parameters when `branch_predictor_name_ != "enhanced"` to fix the regression failures.

core/fetch/Fetch.cpp
```diff
@@ -55,6 +58,24 @@ namespace olympia
         node->getParent()->registerForNotification<bool, Fetch, &Fetch::onROBTerminate_>(
             this, "rob_stopped_notif_channel", false /* ROB maybe not be constructed yet */);
 
+        std::for_each(branch_predictor_name_.begin(), branch_predictor_name_.end(),
+                      [](char& ch) { ch = static_cast<char>(std::tolower(static_cast<unsigned char>(ch))); });
+
```
bdutro, line 63: Alternatively, you can do this here and then use the local variables when you instantiate the EnhancedBranchPredictor:

```suggestion
const auto enhanced_btb_entries = p->enhanced_btb_entries;
const auto enhanced_btb_ways = p->enhanced_btb_ways;
const auto enhanced_bht_entries = p->enhanced_bht_entries;
```

bdutro (approved):

core/fetch/EnhancedBranchPred.cpp
klingaard: Is this still an issue?

bdutro: Oh, good call. @kumavatkomal this should be a right shift.

kumavatkomal: Right shift causes 6 test failures in BranchPred_test. Left shift passes all tests. Should the tests be updated, or am I missing something else that needs to change?

bdutro: Ok, I understand what's happening now. The Sparta cache for the BTB is configured with 1KB lines, so the cache address decoder does `>> 10` to get the index bits. We actually want it to use `PC[8:1]` for the lookup, so left shifting it by 9 bits first takes care of that.

A different way to handle this would be to set `kBTBCacheLineBytes = 2` and configure `btb_cache_` like so:
```
    btb_cache_ = std::make_unique<sparta::cache::SimpleCache2<BTBCacheLine>>(
            btb_entries_ * kBTBCacheLineBytes,
            kBTBCacheLineBytes,
            kBTBCacheLineBytes,
            BTBCacheLine(),
            *btb_replacement_policy_,
            false);
    }
```
This would configure the address decoder in the cache to just `>> 1`, and then this function can be removed.

kumavatkomal: Implemented your suggestion:

- kBTBCacheLineBytes = 2 with btb_entries_ * kBTBCacheLineBytes and false parameter

- Removed btbCacheAddress_() function, now using fetch_pc directly

- Cache decoder now does >> 1 for PC[8:1] indexing

All tests pass. Is this correct?

oluwatimilehin:
Hi all, is there any more work planned for this PR, or is it just waiting to be merged?

klingaard:
Hey Komal, do you have any objection to letting Timi finish this work?

kumavatkomal:
Yes, sure Timi... You can continue with the work. Thanks for checking.

